PR #566, opened 2018-04-11
SolrResult.isDOI does not consider the seriesID

### LINKED ISSUE #479: SolrResult.isDOI does not consider the seriesID
If the DOI is set in the series ID then instead of the PID, `SolrResult.isDOI ` returns False.  Requesting an enhancement to `SolrResult.isDOI ` that also inspects series ID for the DOI.

vchendrix:
This is the fix we are using in ESS-DIVE to determine if a data package already has a DOI.  Thought I would submit the pull request since we already had the solution.

Closes #479

### DIFF
--- a/src/js/models/SolrResult.js
+++ b/src/js/models/SolrResult.js
@@ -176,11 +176,18 @@ define(['jquery', 'underscore', 'backbone'],
 				this.set("url", MetacatUI.appModel.get("resolveServiceUrl") + encodeURIComponent(this.get("id")));
 		},
 
-		// checks if the pid is already a DOI
+		/**
+		 * Checks if the pid  or sid is already a DOI
+		 *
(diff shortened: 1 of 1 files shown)


mbjones:
Looks good.  As an aside, I note that DOI_PREFIXES in that function is incomplete:

```js
var DOI_PREFIXES = ["doi:10.", "http://dx.doi.org/10.", "http://doi.org/10."];
```
and probably should include others, such as:
```js
var DOI_PREFIXES = ["doi:10.", "http://dx.doi.org/10.", "http://doi.org/10.", "http://doi.org/doi:10.", "https://dx.doi.org/10.", "https://doi.org/10.", "https://doi.org/doi:10."];
```

There are quite a few supported format possibilities.

vchendrix:
@mbjones the suggested changes have been made.

PR #592, opened 2018-05-01
Add copy-on-click support to the DOM returned by the view service

### LINKED ISSUE #591: Make alternate data access URLs in view service copiable
As discussed over on https://redmine.dataone.org/issues/8499, we wanted to make the URLs shown in the alternate data access table copiable.

amoeba:
This behavior is slightly different than the Copy Citation in that it just replaces the innerHTML with a ✔ and, after 500ms, returns the old button (or other element) innerHTML. This'll also work if someone needs copiables in other areas of the view service later on.

Preview:

kapture 2018-04-30 at 16 56 31

Closes #591

### DIFF
--- a/src/js/views/MetadataView.js
+++ b/src/js/views/MetadataView.js
@@ -121,6 +121,13 @@ define(['jquery',
 				this.pid = MetacatUI.appModel.get("pid");
 
 			this.listenTo(MetacatUI.appUserModel, "change:loggedIn", this.render);
+
+			// Injects Clipboard objects into DOM elements returned from the View
+			// Service
+			this.on("metadataLoaded", function() {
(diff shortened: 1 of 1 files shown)


laurenwalker:
Nvm, I just figured out how to do it on my end.

laurenwalker:
I moved the call to insertCopiables() into the renderMetadata() function to follow the design of the view, which is to only use render() to set up the basic components of the view and start the fetch the model. All of the detailed elements that are added to the metadata rendering is done in renderMetadata().

I also replaced the checkmark character for the Copy buttons with a FontAwesome checkmark icon to stay consistent with the rest of the app.

I'll be merging this work into the 1.14 branch as well.

PR #593, opened 2018-05-01
Add PANGAEA-specific override to CitationView

amoeba:
This modifies the look of the CitationView when the `datasoruce`
is `urn:node:PANGAEA` and the metadata Object has a SID. When that
is the case, only the SID will be displayed.

@laurenwalker does this conflict with your work on making URIs in the CitationView clickable? 

I did a minor amount of testing and this appears to work as intended.

### DIFF
--- a/src/js/views/CitationView.js
+++ b/src/js/views/CitationView.js
@@ -205,7 +205,13 @@ define(['jquery', 'underscore', 'backbone', 'models/SolrResult'],
 			//The ID
 			var idEl = $(document.createElement("span")).addClass("id");
 			if (seriesId) {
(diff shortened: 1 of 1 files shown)


laurenwalker:
The CitationView has changed since @amoeba created this PR, so I resolved those merge conflicts first. This PR is now good to go. I will need to merge Bryce's change to the 1.14 branch too, since that is our current version installed in production and DataONE needs this as a hot fix.

amoeba:
Thanks a bunch @laurenwalker !

amoeba:
Thanks!

Issue #493, opened 2018-02-21
Issue with additionalinfo de/serializing


vchendrix:
EML xml element `additionalInfo` is not parsing correctly.  An investigation into `EML211.js` shows that a local variable `emlText` (line 335) shadows the array `emlText` (line 246) which lists the.

A pull request for this is forth coming

PR #635, opened 2018-06-01
Needed to use convertedName for emlText

vchendrix:
ESS-DIVE extends MetacatUI to make use of `<additionalInfo>` entity.  There was a previous pull request that enabled the use of `<additionalInfo>` in `emlText`.  See Issue #493.   On `EML211.parse()` it was discovered that `additionalinfo` was not using `nodeNameMap()` to convert to the camel-case entity name (`additionalInfo`)

The changes are as follows:
- additionalInfo was not being converted to camel-case (`additionalinfo > additionalInfo`)
- moved local variable convertedName up to beginning of loop for reuse.


laurenwalker:
Good catch, thanks!

PR #650, opened 2018-06-13
Fix bug in plural form for 'region(s)' in resultItem templates

### LINKED ISSUE #649: SearchResultsView resultItem template uses incorrect plural in tooltip
If I go to https://arcticdata.io/catalog/ I can see this tooltip:

screen shot 2018-06-12 at 3 57 29 pm which uses the plural form of region when it shouldn't. It's also missing a period at the end.

amoeba:
Closes #649

### DIFF
--- a/src/js/templates/resultsItem.html
+++ b/src/js/templates/resultsItem.html
@@ -59,7 +59,7 @@
       	<div class="on-map icons  stop-route">
 		<% 
 		if (typeof southBoundCoord != "undefined"){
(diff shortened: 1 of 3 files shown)


laurenwalker:
Since we decided to not ever delete branches, can we have minor PRs like this done from the target branch rather than a new feature branch?

amoeba:
Ah, whoops -- sorry! I'm in too much of a PR mindset and I keep doing this across projects. Will do next time.

laurenwalker:
Thank you!

PR #706, opened 2018-07-23
Disable download if the package is too large.

rushirajnenuji:
Solved by changing the CSS of that button cursor to default and checking if the button element has the `disabled` attribute.

- The download button works, as expected, for rest of the cases.

### DIFF
--- a/src/js/views/DownloadButtonView.js
+++ b/src/js/views/DownloadButtonView.js
@@ -55,40 +55,43 @@ define(['jquery', 'underscore', 'backbone', 'models/SolrResult'],
 			e.preventDefault();
 						
 			// Checking if the Download All button is disabled because the package is too large
-			var isDownloadDisabled = (this.$el.attr("disabled") == "disabled") ? true : false;
+			var isDownloadDisabled = (this.$el.attr("disabled") === "disabled") ? true : false;
 
-			// Downlading only if the Download button is enabled else do noting.
-			if(!isDownloadDisabled) {
-				if(this.$el.is(".in-progress"))
-					return true;
-				
-				//Show that the download has started
-				this.$el.addClass("in-progress");
-				var buttonHTML = this.$el.html();
-				this.$el.html("Downloading... <i class='icon icon-on-right icon-spinner icon-spin'></i>");
-				
-				//If we found a model, fire the download event
-				this.model.downloadWithCredentials();
-				
-				this.listenTo(this.model, "downloadComplete", function(){
-					
-					//Show that the download is complete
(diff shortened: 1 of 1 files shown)


src/js/views/DownloadButtonView.js
```diff
@@ -45,42 +45,50 @@ define(['jquery', 'underscore', 'backbone', 'models/SolrResult'],
...
-			this.$el.addClass("in-progress");
-			var buttonHTML = this.$el.html();
-			this.$el.html("Downloading... <i class='icon icon-on-right icon-spinner icon-spin'></i>");
-			
-			//If we found a model, fire the download event
-			this.model.downloadWithCredentials();
-			
-			this.listenTo(this.model, "downloadComplete", function(){
+			// Checking if the Download All button is disabled because the package is too large
+			var isDownloadDisabled = (this.$el.attr("disabled") == "disabled") ? true : false;
```
amoeba: It'd be good to use type stable comparison here, `===` (and everywhere, generally-speaking). What do you think about changing this from `==` to `===`?

src/js/views/DownloadButtonView.js
```diff
@@ -45,42 +45,50 @@ define(['jquery', 'underscore', 'backbone', 'models/SolrResult'],
...
+
+			// Downlading only if the Download button is enabled else do noting.
+			if(!isDownloadDisabled) {
```
amoeba: This code looks just fine and makes sense. One thing I like to do to avoid deeper nesting in my code is to follow the early return pattern, which changes code from:

```js
var enabled = true;
var other_condition = true;

if (enabled) {
  if (other_condition) {
	// do x
  }
}
```

which has two levels of `if`'s nested.

to

```js
var enabled = true;
var other_condition = true;

if (!enabled) {
  return;
}

if (other_condition) {
  // do x
}
```

The other conditional in this method looks like it follows the early return pattern. What do you think about changing this around here?

rushirajnenuji:
Hey @amoeba, thanks for the review. Yes, I agree, both these changes are not only technically correct but also, are good SE practices. I've updated the code with the commit https://github.com/NCEAS/metacatui/pull/706/commits/dfcb41c572c0c5d8865ea627b1599d9fa70f61f0

Will definitely keep these points in mind for future development. I appreciate your time and review.

amoeba:
Look good, thanks @rushirajnenuji ! @csjx wanted to look and and deploy this later on tonight so I'll let him merge when he's ready. Thanks for chipping in to get the fix in!

@csjx which branch does this need to land on? dev 2.2? I wasn't sure so I didn't change the PR base.

PR #709, opened 2018-07-24
Add support for Highwire Press bibliographic meta tags

### LINKED ISSUE #491: Add support for bibliographic import tools
@mbjones mentioned today that, at some point in the past, bibliography importing tools such as the one built into Zotero were able to pick up basic bibliography from our dataset landing pages. I'm not sure when or why this broke but I was just able to confirm that it doesn't work with Zotero or Mendeley's Chrome extensions. I think that getting this implemented would (1) not be a lot of work and (2) aid us in our mission to treat datasets in DataONE as first-class, citable resources.

After a bit of research into how these integrations work, it appears to be a total mess as to *how* to do this but it appears very doable. I quickly found some helpful links:

- https://www.zotero.org/support/dev/exposing_metadata
  They support:
  - Embedded Metadata (`<meta>` tags): Dublin Core and Google/Highwire
  - COinS
  - unAPI
- https://scholar.google.com/intl/en/scholar/inclusion.html#indexing
  - Highwire
  - Eprints
  - BE press
  - PRISM
  - Dublin Core, though they note:
  > Use Dublin Core tags (e.g., DC.title) as a last resort - they work poorly for journal papers because Dublin Core doesn't have unambiguous fields for journal title, volume, issue, and page numbers.
- I found the source code for what appears to be Zotero's implementation:
 https://github.com/zotero/translators/blob/b76abc351e5d6c88f09b8159b532cb4415700458/Embedded%20Metadata.js#L40
- https://webmasters.stackexchange.com/questions/11613/indexing-for-google-scholar-which-tags-to-use
- http://web.archive.org/web/20160408141640/https://www.mendeley.com/import/information-for-publishers/

  From Mendeley:
  > There are 4 supported "standards" for describing articles in the META elements, listed in the order of usage preference:
  > - Highwire Press tags (citation_*)
  > - Eprints tags (eprints.*)
  > - PRISM tags (prism.*)
  > - Dublin Core (dc.*/dcterms.*)

It looks like Highwire Press tags are a pretty well-supported standard so maybe they'd be a good start?

To implement this, we should resolve some questions:

1. Can the systems that support these standards pull the information from a client-side webapp?
2. Where would we get the information and what information would we expose?
3. Which standard would we use? Highwire?

I think checking (1) should be pretty easy and would be good to check first as it'd be a dealbreaker if it didn't pan out.

amoeba:
This change required a refactor of the JSONLD work to DRY up code that needed similar logic between the JSONLD and Highwire Press code.

While I was in here, I realized we weren't properly removing stale JSONLD so I added a method that is attached to all routers' 'route' event that clears the stale JSONLD and also the Highwire Press meta tags.

Resolves #491 if we think this is enough for now.

### DIFF
--- a/src/js/routers/router.js
+++ b/src/js/routers/router.js
@@ -47,6 +47,10 @@ function ($, _, Backbone) {
 			this.route(/^view\/(.*)$/, "renderMetadata");
 
 			this.on("route", this.trackHash);
+			
+			// Clear stale JSONLD and meta tags
+			this.on("route", this.clearJSONLD);
+			this.on("route", this.clearHighwirePressMetaTags);
 		},
 
 		//Keep track of navigation movements
@@ -539,6 +543,17 @@ function ($, _, Backbone) {
 				MetacatUI.appView.statsView.onClose();
 			else if(lastRoute == "profile")
 				MetacatUI.appView.userView.onClose();
+		},
+
+		clearJSONLD: function() {
+			$("#jsonld").remove();
+		},
+
+		clearHighwirePressMetaTags: function() {
+			$("head > meta[name='citation_title']").remove()
+			$("head > meta[name='citation_authors']").remove()
+			$("head > meta[name='citation_publisher']").remove()
+			$("head > meta[name='citation_date']").remove()
 		}
 
 	});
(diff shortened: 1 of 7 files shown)


amoeba:
Just re-sync'd this with dev-2.1. Can I get a +1 and a merge @laurenwalker or @rushirajnenuji ?

PR #839, opened 2018-11-20
Add a toggle-able dropdown to the metadata view that allows someone t…

ThomasThelen:
# Whole Tale Environments

This PR is for a feature that allows a user to come from DataONE to Whole Tale with the intent on using a data package. To do this I added a dropdown menu next to the copy button in the metadata view.

## Whole Tale Flag

I added a flag to the appmodels so we can easily toggle the feature on and off. You can see where the application uses this on line 955 in `src/js/views/MetadataView.js`

## Dropdown Menu

I added the dropdown menu to `src/js/templates/metadataControls.html`. It uses the result from above to check if it should render the content.

## Links For Each Menu Item

I construct the links for each compute environment at line 1032 `src/js/views/MetadataView.js`


ThomasThelen:
I had branched off of master instead of the dev branch. I'm closing this while I rebase and change the merge base.

PR #1034, opened 2019-07-30
Add EML 2.2 support

amoeba:
(Disregard the branch name, I threw more stuff into the single branch for simplicity's sake)

(@laurenwalker 	commented that we will likely target this for 2.7 so, after review, we'll wanna rebase this before merging)

**Adds support for new stuff we have going on in EML 2.2.0:**

- Semantic annotations
  - Arctic theme now has an Annotation filter which I re-used from the previous D1 semantics work
  - MetadataViews on EML 2.2.0 docs with `<annnotation>` elements have an interactable popover that shows more details about the semantic annotation and allows the user to start a search from the popover for similarly-annotated records
  - Annotation lookup on landing pages is driven by Bioportal's API and are cached app-wide to avoid re-requesting data from Bioportal
- Markdown support in `TextType` elements
  - MetadataView now creates `MarkdownView`s for every `.markdown` element it sees on the page. 

To test, you'l need to point your MetacatUI instance at a `baseUrl` of `https://neutral-cat.nceas.ucsb.edu` and set your theme to `arctic`.

### DIFF
--- a/src/components/bioportal/jquery.ncbo.autocomplete-2.0.1.js
+++ b/src/components/bioportal/jquery.ncbo.autocomplete-2.0.1.js
@@ -186,22 +186,6 @@
             text = text.replace(filterPatt, '<strong>$1</strong>');
           }
 
-          // Add in match type and any synonyms
-          var matchTypeTag = "";
-          switch(resultObjects[i].matchType) {
-            case "prefLabel":
-              matchTypeTag = '<span class="ncboAutocomplete-matchType-label"><abbr title="Matched by label">LBL</abbr></span>';
-              break;
-            case "synonym":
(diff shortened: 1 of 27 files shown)


amoeba:
I need to make one small tweak but otherwise this is ready for review. Making that change now.

laurenwalker:
Thanks for the PR, Bryce. I saw the discussion on Slack that the release of EML 2.2.0 is on the way and that we are going to discuss all your changes to MetacatUI and Metacat with the larger DataONE group. I'll wait until we have that discussion and until the 2.2.0 release before I merge this PR in.

amoeba:
Sounds good, thanks.

amoeba:
Okay this is good for review, @laurenwalker 	. Made the changes we talked about on today's dev call.

PR #1149, opened 2019-10-21
Adds updating sysmetadata function to DataONEObject file.

### LINKED ISSUE #1148: Add the ability to update metadata
To be able to update metadata objects, we need to add an `updateSysMeta()` function that update the metadata through the `/meta` endpoint.

helbashandy:
Update includes:
	- Pointing to /meta endpoint for the added `updateSysMeta()` function.
	- systemMetadataOnly option is passed to serializeSysMeta function to be able to ignore the null setting of checksum.

Closes #1148

### DIFF
--- a/src/js/models/DataONEObject.js
+++ b/src/js/models/DataONEObject.js
@@ -625,6 +625,109 @@ define(['jquery', 'underscore', 'backbone', 'uuid', 'he', 'collections/AccessPol
               //Send the Save request
               Backbone.Model.prototype.save.call(this, null, requestSettings);
         },
+        /*
+          * Updates the DataONEObject System Metadata to the server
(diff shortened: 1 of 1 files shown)


laurenwalker:
I changed the code merged from this PR since it was calling Backbone.Model.save(), which we don't want to do. The callback functions also had a lot of save-specific event triggering, which I removed. I created a "sysMetaUpdated" event for the app to listen to instead, for only listening to the successful system metadata saves.

helbashandy:
@laurenwalker  I was not sure which events were save specific so I removed those that I was sure didn't relate and left those that I thought would need a second eye from you. Thank you for looking into it and merging the PR!

PR #1319, opened 2020-03-26
[WIP] Update Readme to use sites-available configuration

ThomasThelen:
### Summary

As per our discussion on Slack, we want to standardize the apache configuration for metacatui in the `/etc/apache2/sites-available` directory. 

This PR removes references to the virtual hosts file and the httpd config file. They are replaced with a template that the user can place in their `sites-available` folder with a few examples.

The easies way to look at this is probably on [my fork](https://github.com/thomasthelen/metacatui).

### WIP Tag
Before merging, I'd like to make sure I can get it to work with the metacat.lcoal settings to make sure I didn't break the guide. I'll assign reviewers when I complete the testing.

To Do:
- [ ] Test that the apache serving portion works when followed

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,51 +28,58 @@ A metadata view, as configured for the Arctic Data Center:
 To run MetacatUI, you will need to first install a web server such as [Apache](https://httpd.apache.org/). The following instructions are for Mac OS X. Apache comes pre-installed on Mac OS X.
 
 #### Step 1. Set up Apache
-- Choose a location from which to serve *all* your Apache website files. A good location is `/Users/{username}/Sites`
-- Make a subdirectory in `~/Sites` specifically for MetacatUI. The default directory name for MetacatUI is `metacatui`.
+
+##### Choosing a location for MetacatUI
+Choose a location from which to serve *all* your Apache website files. A good location is `/Users/{username}/Sites`
+
+Make a subdirectory in `~/Sites` specifically for MetacatUI. The default directory name for MetacatUI is `metacatui`.
 
   ```
   mkdir ~/Sites/metacatui
   ```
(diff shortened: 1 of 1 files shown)


laurenwalker:
Hi @ThomasThelen - Can you confirm that your updated Apache instructions worked on Mac OS? I tried following your instructions and I don't have a /sites-available directory in my local Apache. So I think there is a missing step somewhere, or I'm forgetting something.

ThomasThelen:
@laurenwalker I did this on Ubuntu Server. It looks like mac keeps it in a different location.
I was running metacatui today in the httpd Docker container (which awesomely worked) and noticed that it too had a different directory layout than my Ubuntu Server.

Testing this on a fresh Ubuntu Server and macOS will take me a bit of time and I'm likely to lag a little here until later in the week. I'll can push my changes back here and ping you on slack after! My solution is to just split the installation instructions into...

### Installation

#### With Local Metacat
Keep the same

#### macOS / OS X
Current readme

#### Ubuntu Server
My readme

#### Configuring MetacatUI
Since this content is independent of Apache configuration stuff, make a small section for it. 

##### Setting Themes

##### Configuring Metacatui Paths
Stuff like `data-metacat-context` and friends

##### Configuring Coordinating Node Locations
Blurb about setting the CN info in AppModel

laurenwalker:
Okay, sounds great. I agree about having separate instructions for Ubuntu and MacOS. I am thinking these detailed instructions should actually be in a different markdown file, linked to from the README, to avoid clutter. But I can organize that later.

laurenwalker:
We've moved the install instructions to a different doc location, and have Apache steps already complete

Issue #521, opened 2018-03-09
add ability to reorder EMLParties


mbjones:
Creator order is preserved in the citation.  Need to be able to reorder creators easily.  Probably easiest with drag and drop on the rows with a drag handle.  But could also be done with Move up/Move down arrow butttons.

vchendrix:
We just got this feature request from one of our users.

laurenwalker:
Thanks, Val. I'm going to incorporate this feature into a release soon.

laurenwalker:
To start, this feature should use a simple move up/down buttons, rather than drag-and-drop, since the EMLPartyView tends to be pretty tall. 

When designing the Portal Editor, we came up with a redesign of the EMLPartyView that would make drag-and-drop reordering much simpler. When we redesign the EMLPartyView, we can introduce drag-and-drop reordering at the same time. I created a new issue to contain that iterative enhancement: #1239

amoeba:
Partial support pushed up in https://github.com/NCEAS/metacatui/commit/64ea4f67d18509f7f5f08ee23c037c0b56f46817. Will continue on this tomorrow AM. You can currently move a party down and it moves the party down the page and in the serialized EML doc. Want to find a nice way to display the move buttons and catch a few edge cases.

amoeba:
Simple up down button implementation PR'd in #1334. 

Action shot:

![Kapture 2020-04-16 at 16 11 02](https://user-images.githubusercontent.com/563/79518193-ecd3a180-7ffc-11ea-8f9f-5138f19abdee.gif)

Issue #1238, opened 2020-01-13
Redesign EMLPartyView


laurenwalker:
When we were designing the Portal Editor, we redesigned the EMLPartyView as well. It fixes some usability issues and allows for easier drag-n-drop reordering of EMLParties.

See the redesign mockups here: https://github.nceas.ucsb.edu/NCEAS/nceas-design/tree/master/MetacatUI/Portals/Portal-Editor/2019-06-05_Round2/2x

PR #1334, opened 2020-04-17
Add the ability to re-order parties

amoeba:
Closes https://github.com/NCEAS/metacatui/issues/521

As decided in #521, the initial implementation adds simple up/down buttons alongside each party view. Clicking either button moves the party up or down on the page and also updates the underlying EML model so the serialized document reflects the change. When re-ordering creators, the citation preview automatically updates to reflect the changes.

Three things I'd appreciate an extra look at are:

1. I don't let you move a party down below the new/placeholder party so the placeholder is always the last item in the group. It made sense to me to leave the placeholder at the bottom.
2. I didn't wanna to get too fancy with the up/down buttons so I just put them to the left of the party menu. Very much open to feedback here.
3. I wanted to disable the up or down button when it would have no effect (i.e., the Up button would be disabled on the first item) but was having trouble finding a clean way to do it. Right now, you can click the button but it does nothing.

### DIFF
--- a/src/js/models/metadata/eml211/EML211.js
+++ b/src/js/models/metadata/eml211/EML211.js
@@ -1782,6 +1782,44 @@ define(['jquery', 'underscore', 'backbone', 'uuid',
        *
        * @param {EMLParty} partyModel: The EMLParty model we're moving
        */
+      movePartyUp: function(partyModel) {
+        var possibleAttr = ["creator", "contact", "metadataProvider", "publisher", "associatedParty"];
+
+        // Iterate over each possible attribute
+        _.each(possibleAttr, function(attr){
+          if (!_.contains(this.get(attr), partyModel)) {
+            return;
+          }
+          // Make a clone because we're going to use splice
+          var models = _.clone(this.get(attr));
+
+          // Find the index of the model we're moving
+          var index = _.findIndex(models, function(m) {
+            return m === partyModel;
+          });
+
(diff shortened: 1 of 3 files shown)


mbjones:
Awesome, @amoeba.  Can you paste a screenshot in here so we can see the UI layout in action?

amoeba:
Sure:

Here's me reversing the order of three creators:

Kapture 2020-04-16 at 16 25 38

laurenwalker:
Looks great to me, Bryce!! About your 3 items for review: I think that all sounds fine. I was originally imagining the Move Up and Move Down buttons to be options inside the menu next to each party (the menu is accessible via the three vertical dots). But since we have plans to eventually redesign the EMLPartyView anyway (#1238), I think we can keep it as-is for now, unless that's a super quick fix.

Thanks again!

amoeba:
> I was originally imagining the Move Up and Move Down buttons to be options inside the menu next to each party...

Ah, okay. I assumed you wanted literal buttons when you said,

> To start, this feature should use a simple move up/down buttons

I like the buttons being visible without being behind a click since I actually didn't know the three dot button on the right was there until the other day. Putting them there is a super easy change though. Lemme know.

laurenwalker:
I think it's fine as-is

mbjones:
I prefer it as is, where the buttons are more evident.

PR #1428, opened 2020-06-08
Fixes checkboxes filters alignment

helbashandy:
This is a small css change that was breaking the alignment for checkboxes filters.


laurenwalker:
Thanks Hesham!

PR #1527, opened 2020-09-16
improvements to KNB about page

jeanetteclark:
This has been sitting around for a while, but some time ago I made some additions to try and help our CoreTrustSeal application. I also think these changes might mitigate some of the issues we have with users trying to use Morpho, because it highlights the web form as the primary submission method.

IIRC Lauren said this was okay to PR against the master branch since it doesn't affect the application but let me know if I should switch to another branch.


mbjones:
The morpho changes on tools look good, but I think you should boldface the phrase: "**This software is no longer maintained by our development team.**" It would also be good to indicate that the MetacatUI web editor is our preferred and maintained replacement. On the morpho README, I added the [Project Status: Inactive – The project has reached a stable, usable state but is no longer being actively developed; support/maintenance will be provided as time allows.](https://www.repostatus.org/#inactive) badge. Maybe that belongs here too?

laurenwalker:
I merged this into the dev-2.13 branch and it will go in the 2.13.3 patch release soon. Feel free to make any more changes and a  new PR. Thanks Jeanette

PR #1535, opened 2020-10-05
Selectively remove alerts on validation

### LINKED ISSUE #1534: temporary message alert gets removed on validation.
**Describe the bug**
After using the temporary message configuration to show a temporary alert as a site-wide notice, I noticed that the alert gets removed after a data package gets submitted on it's own. 

The issue is that when this happens, the `has-temporary-message` class doesn't get removed so it messes up few css classes that depends on the css class to create margins/paddings. 

It looks like the issue is on the `showValidation` function on [this line](https://github.com/NCEAS/metacatui/blob/master/src/js/views/metadata/EML211EditorView.js#L861) specifically as it doesn't selectively exclude the temporary message. 

**To Reproduce**
1. Go to data-dev.ess-dive.lbl.gov
2. Try to submit a package
3. Notice that the temporary message is removed and the class `has-temporary-message` still exist.

**Expected behavior**
I believe one of the two behaviors could be fine:

1. Keep the temporary message at all times after submission (with error and without)
2. Remove the temporary message but also remove the `has-temporary-message` class.

 - OS: MacOs
 - Browser: Chrome
 - Version: 2.12.0

helbashandy:
In order to not remove the temporary message when removing all validation alerts, a selective css selector is used to exclude alerts with temporary message.

closes #1534

### DIFF
--- a/src/js/views/metadata/EML211EditorView.js
+++ b/src/js/views/metadata/EML211EditorView.js
@@ -858,7 +858,7 @@ define(['underscore',
           this.$(".notification.error").empty();
           this.$(".side-nav-item .icon").hide();
           this.$("#metadata-container .error").removeClass("error");
-          $(".alert-container").remove();
+          $(".alert-container:not(:has(.temporary-message))").remove();
 
 
           var errors = this.model.validationError;


laurenwalker:
Thanks Hesham!

Issue #1403, opened 2020-05-22
support structured funding info from EML 2.2.0


mbjones:
**Describe the feature you'd like**
Structured funding information was added to EML 2.2.0: https://eml.ecoinformatics.org/whats-new-in-eml-2-2-0.html#structured-funding-information We should support editing that more detailed information.

- [ ] Display structured funding info
- [ ] Edit structured funding info

**Is your feature request related to a problem? Please describe.**
Its difficult to machine-parse our current natural language funding information.

**Additional context**
- If `funding` information present in a 2.1.1 or earlier document, we should make an attempt to forward convert that to the new EML 2.2.0 `award` element structure if possible using common patterns for award identifiers.
- Funder identifiers might be able to be looked up from the FundRef system by CrossRef, or maybe through [ROR](https://ror.org), or others.

laijasmine:
We run into an issue now with editing EML 2.2.0 documents when we use the editor. The awards field will be empty if we are using the new structured funding info and you can't save until you add something into that field because the field is required. So we will need to do an extra step in R to remove the funding information added every time we use the editor.

mbjones:
@amoeba is this a side effect of simply serializing into EML 2.2.0 format? The new `award` field is optional and so shouldn't be present in any autoconverted document unless specific steps are taken to transform `funding` field info into the `award` subfields. Thoughts?

amoeba:
Sounds like we missed this in the switch to EML 2.2.0: I think we just need to change MetacatUI's logic for checking whether funding information has been set. Currently, `dataset/project/funding` is probably required but we should really check for `dataset/project/funding` _and/or_ `dataset/project/award`, considering having at least one as "having funding".

amoeba:
After looking, I ran into a question: Do we have a use case for deployments wanting to accept `funding` **or** `award` or is our plan to make the funding portion of the editor automatically migrate `funding` into an `award` element on parse and _only_ allow editing of funding information using the `award` element?

mbjones:
The way we redefined `funding`,  it becomes more of a natural language description and can coexist with the structured funding info in `award`. I could see having a paragraph in `funding` along with one or more `award` entries. That said, if all `funding` contains is something like `National Science Foundation 198765` than I think it would be better to convert that into the `award` structure.

amoeba:
Gotcha. I think allowing editing of the old `funding` field will probably be useful to someone.

[...]

Issue #814, opened 2018-10-29
Project section pre-populates incorrect metadata


dmullen17:
@jagoldstein and I noticed that the `title` and `personnel` sections default to the title of the submission and the name of the creator.  It would be preferable if the editor populated the project title section with something generic like: `NSF-funded project`.  We will normally update this section internally, and in the event that we fail to do so the title will be generic rather than incorrect.

laurenwalker:
Related to #525 - Add Project Section to the Editor. This will allow the project metadata to be edited in the web editor.

The title and personnel are both required fields in the EML project module, so unfortunately the editor cannot complete omit those.

jagoldstein:
It sounds like a generic project title would be a good interim solution. `NSF-funded project` or even`NA`would be satisfactory.

The personnel field should probably be populated similarly until the Project Section of the Editor is released. `submitter`s and even `creator`s are not always the project PI. A value of `PI` or `NA` should be okay defaults to populate that field with for now.

Issue #525, opened 2018-03-13
Add Project section to the Editor


laurenwalker:
We need to add the ability to edit the project personnel, first and foremost, since it is automatically created with the first Creator info.

Adding other project metadata fields will be useful too, e.g. abstract. But that can be done in a future stage, for time's sake.

laurenwalker:
A couple weeks ago at our weekly dev call we discussed this and decided that we should add a project section to the editor sooner rather than later. Since we insert the funding information into the project section, the personnel and project title are automatically filled in behind the scenes. This metadata should be editable in the web UI. Otherwise, users will need to resort to manually editing the EML to change the information if it is incorrect.

Assigning to 3.1.0 for now

laurenwalker:
This is something that our Arctic data team still does a lot manually via R. And it will likely become an issue in the future for our hosted repos as they start inputting funding information via the editor (that's when the project sections are created). Moving up to a near milestone

laurenwalker:
I spoke with @wykhuh about starting on this project for her ADC data fellowship. 

I made this issue into an Epic on Zenhub and added three issues that are sub-tasks. By the end of her fellowship, we'd like to have a ready-to-release version of the Project section, even if it is just editable fields for the structured funding metadata (see #1403) although ideally the project `title` and PIs would be editable, too (see #814).

**For a bare-bones starter task, we decided on:**

- [ ] Create a new Backbone View, called `EMLProjectView`, and a "Project" editor section. When the "Project" section title is clicked, the `EMLProjectView` is rendered, with a simple "Hello World" message

Wai-Yin will keep the team updated on her progress by frequent commits to her MetacatUI fork, comments here in Github, and Slack.

PR #1632, opened 2020-12-18
Feature/add project section

wykhuh:


laurenwalker (commented):
Looks great so far. I'm merging into a feature branch until the rest of the development is done and it's ready for release.

Related to #525 #1403 #814

laurenwalker:
This looks really great so far and is a great start, Wai-Yin. Here's some more detailed feedback:

## Detailed review

Screen Shot 2020-12-18 at 10 33 54 AM

- [ ] Make form labels bold, to match the other labels in the editor.
- [ ] Add help text under each label that explains what that field means - We can reuse text from the [EML 2.2.0 Award documentation](https://eml.ecoinformatics.org/schema/eml-project_xsd.html#ResearchProjectType_award) to start with. This help text should be styled in the same way it is elsewhere in the editor (the subtle grey text, on a separate line).
- [ ] Also add help text for the Project section in general, under the top Project header
- [ ] Add an "Awards" header and help text for awards in general
- [ ] Let's change "Title" to "Award title", to be more specific
- [ ] Perhaps change the order of the fields? Group the award and funder fields together
- [ ] For the Arctic Data Center, we want to lookup awards via the NSF Awards API and pre-populate these fields. We already do this in the current Funding field in the Overview section, so let's move that functionality here. We should indicate in the help text, like we do now:

Screen Shot 2020-12-18 at 10 43 33 AM
- [ ] MetacatUI can be confiured to not use the NSF AWard API, because not all repos that use MetacatUI have such a direct tie to NSF like the ADC does. So we only want to show help text about the NSF Award lookup or perform the lookup if [`useNSFAwardAPI` is set to `true` in the `AppConfig`](https://nceas.github.io/metacatui/docs/AppConfig.html#useNSFAwardAPI) (`MetacatUI.appModel.get("useNSFAwardAPI")`)
- [ ] Hovering over or focusing on the Remove button for an award should show a tooltip that says "Remove this award". We use the Bootstrap tooltips
- [ ] Let's add some vertical margin in between the awards.
- [ ] We should add global configurations for all these field labels since every repository may use slightly different vernacular (e.g. "Award Identifier" instead of "Award Number"). 
- [ ] And each field should be able to be enabled/disabled in the global configuration, too, as long as they are not required by the EML Project element. The optional fields are funderIdentifier, awardNumber, and awardUrl, so those should be able to be disabled in the global AppConfig. We could create a new config like this:

 ```js
emlEditorAwardFieldLabels: { 
funderName: "Funder name",
funderIdentifier: "Funder identifier", 
awardNumber: "Award number", 
awardUrl: "Award URL",
title: "Award title"
}

emlEditorAwardFields: { 
funderIdentifier: true, 
awardNumber: true, 
awardUrl: true,
}
```

The first object maps EML field names to a label and the second controls which fields are displayed in the editor. We previously relied on themes to override templates to customize display and functionality like this, but we've found it makes MetacatUI maintenance and upgrades more problematic.
- [ ] Lastly, add JSDoc documentation to all of the functions and attributes in `EMLProjectView`

laurenwalker:
Another few things:

- [ ] [Do not update the `changed` attribute on the root data package from the EMLProjectView](https://github.com/NCEAS/metacatui/blob/de9ff801fd64404debfa450ac435149f90567686/src/js/views/metadata/EMLProjectView.js#L87-L88). The EMLProject model already listens to changes on it's attributes and will "trickle up" the change to the root data package. We probably just need to start listening to awards: https://github.com/NCEAS/metacatui/blob/de9ff801fd64404debfa450ac435149f90567686/src/js/models/metadata/eml211/EMLProject.js#L33-L36
- [ ] Let's avoid manipulating the EMLProject model too heavily in the views. Let's create a `EMLProject.removeAward()` function that does this:
https://github.com/NCEAS/metacatui/blob/de9ff801fd64404debfa450ac435149f90567686/src/js/views/metadata/EMLProjectView.js#L72-L79
instead of doing that in the view.

Issue #1659, opened 2021-02-04
Tweak BioPortal tree widget behavior to require fewer clicks


amoeba:
The BioPortal tree widget we use in the search catalog and are soon going to use in #1402 and #1653 has a behavior I think most of us find odd: When you type into the autocomplete box and select an item by clicking on it, the tree below expands and shows you your selection. To actually pick that item in our search, you have to then go click on the item in the tree. You can't just pick an item from the autocomplete.

While working on #1402, I added a configuration option to change this behavior so you can just click and be done:

![2021-02-04 12 39 23](https://user-images.githubusercontent.com/563/106959881-c7c53200-66e7-11eb-8d6b-9c49c6f5db3b.gif)

I'll note that you can still use the tree to pick items. I got a few thumbs up in Slack on this idea but would love thoughts or 👍. I think if we like this change, we can make it the default behavior of the widget and make all instances of the widget use the default behavior.

PS: I think this change makes the two buttons on the right pointless:

<img width="344" alt="Screen Shot 2021-02-04 at 12 55 17 PM" src="https://user-images.githubusercontent.com/563/106960274-4cb04b80-66e8-11eb-95ad-c296d79dd259.png">

So they could be absent when the widget is configured in this way.

laijasmine:
@amoeba one clarifying question because I couldn't really tell in the gif - can the tree "open up" to the term after you have made the selection? Kind of like this where it highlights your selected term -

<img width="446" alt="Screen Shot 2021-02-04 at 2 17 09 PM" src="https://user-images.githubusercontent.com/13112379/106962312-ca2d8900-66f3-11eb-80bf-731925abd515.png">

It might help guide people to more specific and accurate terms. When I use the Bioportal interface, I often find myself looking at the tree after finding an annotation just to make sure that my selection of the semantic term makes sense by looking at what it is nested under. Or if it could be more specific by looking at any terms under it.

amoeba:
Yeah I think we could do that and that it could be useful. 

>  just to make sure that my selection of the semantic term makes sense by looking at what it is nested under

btw, this is really good feedback.

amoeba:
hey @laijasmine, so I put together a demo which you can play with at https://neutral-cat.nceas.ucsb.edu. Let me know what you think about how that works.

mbjones:
Worked great for me on a quick run through. 👍

yvanlebras:
[...]

Issue #1402, opened 2020-05-21
support creating and editing EML 2.2.0 annotations


mbjones:
**Describe the feature you'd like**

EML 2.2.0 added the eml-semantics model, which we need to support in the editor. See https://eml.ecoinformatics.org/whats-new-in-eml-2-2-0.html#supporting-semantic-metadata We should support at least two major types of annotations (in priority order):

1. annotations on eml attributes for measurement types
2. general dataset-level annotations

**Is your feature request related to a problem? Please describe.**
We currently lack the ability to edit attribute annotations and dataset annotations, which means the work on our semantic search is at a standstill, and it causes curation bottlenecks.

**Additional context**

Dataset-level annotations are very closely tied to controlled keywords in the EML `keywordSet` element with a thesaurus, and should be harmonized. Its probably best to use `keywordSet` for uncontrolled keywords, and `annotation` for controlled keywords if the vocabulary can uniquely identify terms in URI-space.

Other annotation types are possible in EML documents, but our first pass at editing in MetacatUI might simply enable these two types of annotations, and be sure to not lose the others that are found in a document. 

One challenge will be in handling the multiple locations in which annotations can be stored in an EML document, and the proper reference resolution for those in the general `annotations` element and in `additionalMetadata`.

jeanetteclark:
Given our project to assign Arctic Data Center datasets a theme based on a custom vocabulary, it would be good to enable support for a user to assign their own dataset (via annotating the dataset element) a theme based on a custom list. A dropdown menu or lookahead would be really cool.

amoeba:
This is coming along well. I have dataset- and attribute-level annotations parsing, serializing, and editable via the web interface. I'm trying to track down some bugs in how we update and show inline validation errors and I think I'm close there. @mpsaloha and I will be discussing the UI/UX of this today and I'll want to share our thoughts with the group for discussion.

laurenwalker:
Thanks Bryce, those mockups look good. I really like the Measurement Type mockup and don't really have any feedback on it because it's pretty clear as-is. I do think it might make more sense to display it below the Attribute Label input, so that Attribute Name and Attribute Label can be right next to each other. 

Will we also be able to make inferences of the measurement domain based on the annotation added? (e.g. if someone chooses a carbon flux annotation, we could disable the "date-time" measurement domain, at least.)

I'm not entirely sure on the context of the "Annotations" mockup(s). Where will those be shown in the editor?

amoeba:
[...]

PR #1660, opened 2021-02-05
Add annotation support to the EML editor

amoeba:
Implements https://github.com/NCEAS/metacatui/issues/1402.

- Adds support for parsing and serializing dataset and attribute-level semantic annotations.
- Creates a MeasurementTypeView that lets the user pick a single term to create an annotation by using a pre-determined property. Hopefully this makes adding annotations much more approachable.
- Extends BioPortal tree widget to configure behavior when clicking on autocomplete results See #1659.
- This includes some CSS/HTML tweaks to the Entity and Attribute views so they play nice with the MeasurementTypeView.

This branch is deployed at https://neutral-cat.nceas.ucsb.edu for testing.


samanthacsik:
@amoeba, @mpsaloha and I spend some time playing around with the annotation editor and have some comments/suggestions that I'll summarize below (some of these may be a bit redundant from prior conversations, but figured it'd be best to put them all in the same spot). Please also let me know if/where to create separate issues, if that's easier for you to keep track of things.

1. It might need to be clearer that we want a researcher to describe _all_ attributes from their data file (we can see how one might choose to skip/omit attributes, particularly when they have a dataset with many columns). Some suggestions:

- Refine the description under **Attribute** to explicitly say _"Describe **all** the attributes (i.e., fields or variables)..."_
- Auto-add the attributes of a data file (or at least .csv files) so that the researcher only needs to click down the list of automatically-added attribute names and complete the remaining description fields (i.e. label, measurement type, definition, category)
- _Alternatively_ you might have a parsed version of the data file displayed (or a clickable tab which displays the header/first few data rows) so that the researcher can at least see their attribute names in the file without having to pan back and forth between Excel and their browser to ensure that the attribute names match up.
- If parsing the data file is possible, displaying a message next to the `Done` button that lets a researcher know how many attributes they have left to add/describe (e.g. "You've successfully added 6 of 10 attributes! Please add and describe the remaining 4 attributes")

2. Can we add a feature that allows a user to reorder (e.g. drag and drop to rearrange) attributes as they add them? I could be wrong about this, but I assumed that the order of attributes as they appear on the UI should match the order that they appear in the data file? It can be easy to either miss an attribute or add them in the wrong order so an easy way to rearrange them might save some headache (for at least the datateam who's likely to be checking over things).

3. When searching for a `Measurement Type` it would be nice if the page scrolled you down to the area of the hierarchy where your search term is found. For example, if you search for `lower bound soil depth` the user has to scroll a ways down to actually see that section of the expanded hierarchy. For someone who's unfamiliar with ontologies/semantics, they might not even think to scroll down to see where their search term fits within the hierarchy and/or expand to see any potentially relevant subclasses (which may ultimately be a more appropriate annotation). Some suggestions: 

- When someone searches for and selects a measurement type, _only show_ the selected term (along with the `+` for subclass expansion), sibling classes, and parent class. For example, if I search for `lower bound soil depth` only display this part of the hierarchy:
Screen Shot 2021-03-30 at 10 07 06 PM

- A separate but related suggestion: don't automatically add an annotation for the term selected in the Measurement Type search field, but rather, when a researcher searches/selects a term from the search field suggestions, just direct them to the location of that term in the hierarchy. To actually add an annotation, they would have to then click on the term in the hierarchy itself. This would hopefully encourage researchers to expand subclasses to find a potentially more appropriate term, rather than just stopping once their search term has been auto-added as an annotation. An example that comes to mind is if someone searches for `carbon dioxide flux`, they won't see `stomatal conductance` appear as an option in the search field suggestions (even if that may be the best annotation choice). By auto-adding the `carbon dioxide flux` annotation, they might not realize that there are more refined subclasses to choose from.

4. The `Done` button could be grayed out until all required fields have been added so that it's clear a user is missing information before getting all the way back to the home screen and seeing the red "!" indicating missing fields. Though as I type this out, I'm realizing that many users will probably not completely describe their data in a single session but rather save and return later on. So maybe worth keeping as-is...

5. If a researcher chooses a semantic annotation that has a definition and/or description associated with it, it might save them time if we auto-populate the required "Definition" field with that definition/description from the ontology. Then, the "Definition" field may only be required if the researcher wants to add/modify the auto-populated definition. One might click on a "No modifications" checkbox if they don't have a need to add different/additional text.

6. We found a small bug when modifying your term choice from the Measurement Type tree. If you modify your selection and click on a Sibling or Superclass, it highlights your choice in the hierarchy and updates the annotation, but if you choose a subclass (of the original selection), it won't highlight the term you're clicking on. It _does_ however still update the annotation. See screenshots below:

a. My original selection was `Carbon Flux` -- that term is now highlighted in the hierarchy
Screen Shot 2021-03-30 at 10 59 15 PM

b1. I decide to change my selection to Superclass `Flux Measurement Type` -- notice that it highlights that term in the hierarchy + changes the semantic annotation
Screen Shot 2021-03-30 at 10 59 25 PM

b2. I can also change my selection to Sibling class `luminous flux` and it will highlight/update the annotation
Screen Shot 2021-03-30 at 10 59 41 PM

c. I decide to change my selection to a subclass `Carbon Dioxide Flux` -- notice that doesn't highlight my selection, though it _does_ update the semantic annotation
Screen Shot 2021-03-30 at 10 59 52 PM

amoeba:
Thanks very much @samanthacsik for writing this up and to both you and @mpsaloha for the test drive. A few of the comments are out of scope for the work here but I can move them over to the issue we're using to track those kinds of changes.

The two actionable items from above I see are:

- Consider scrolling on behalf of the user to put the tree widget more completely in view after the user selects a term from the autocomplete

  I think this is a good bit of feedback. The tree widget is really awkward as it is and I think this would improve things
- Fix the behavior where some classes aren't selected in the tree after a click, whereas some are

  Great catch, I hadn't noticed this. It's more widespread than you mention and I'm not sure what could be causing it. Probably something simple to fix.

I think these should be quick and I'll incorporate them into the first release here shortly.

amoeba:
Hey @laurenwalker, this is ready to review if you like or merge otherwise. I've tested this a lot and had a few other folks test it too.

To see it in action, either

1. Go to https://neutral-cat.nceas.ucsb.edu
  a. Create a new dataset or edit an existing one
  b. Add/edit an entity and go to the attributes pane
  c. Add/edit an attribute as needed
  d. Scroll down to the Measurement Type section

Or, if setting up locally, do the steps in (1) but ensure your config has:

  ```
  bioportalAPIKey: "{YOUR_KEY}",
  enableMeasurementTypeView: true
  ```

The MeasurementTypeView is feature-flagged and is also hidden if the BioPortal API key is not set. The view allows the user to add one (or many, if configured) EML semantic annotation to an each attribute using an autocomplete+tree viewer widget we use elsewhere. The view only manages semantic annotations with the predicate `containsMeasurementsOfType` and all other annotations are hidden but otherwise unmodified.

Some things I could use feedback if you're interested are:

- Does the placement within the attribute view work with you?
- Do you think my explanatory text is reasonable enough?
- Clicking an item in the autocomplete list scrolls the tree to that item. From there the idea is that you'd pick the term you just went to or browse around up or down the hierarchy. Does this feel pretty good to you?

amoeba:
Heya @laurenwalker, I re-merged develop onto this and tested to make sure the changes still work as expected and things look good. Do you think you'd have time to review this for an upcoming release? 

PS: The neutral-cat deployment linked above isn't live now because I need to test another branch there but a config that has the following two keys (plus the usual ones) is what you need :

```
  bioportalAPIKey: "{YOUR_KEY}",
  enableMeasurementTypeView: true
```

mbjones:
👍 I would appreciate this too -- this work is a critical deliverable for the annotation work on the Data Task Force grant, and we could really use it in production on KNB and ADC. Thanks @laurenwalker for getting it in ASAP.

laurenwalker:
I tried testing this branch today and did not see the annotation editor. There was an error in the console as I was adding attributes:

```
Error saving draft: TypeError: measurementScale.updateDOM is not a function
    updateDOM EMLAttribute.js:368
    updateDOM EMLEntity.js:433
    Underscore 2
    updateDOM EMLEntity.js:432
    updateDOM EMLOtherEntity.js:121
    serialize EML211.js:1046
    Underscore 2
    serialize EML211.js:1035
    saveDraft EMLEntityView.js:566
    jQuery 7
    Backbone 6
    showEntity EML211EditorView.js:960
    jQuery 7
    Backbone 2
    render EditorView.js:81
    render EML211EditorView.js:124
    showView AppView.js:182
    renderEditor router.js:307
    execCb require.js:1635
    check require.js:871
    enable require.js:1112
    bind require.js:129
    emit require.js:1155
    each require.js:57
    emit require.js:1154
    check require.js:925
    enable require.js:1112
    bind require.js:129
    emit require.js:1155
    each require.js:57
    emit require.js:1154
    check require.js:925
    enable require.js:1112
    bind require.js:129
    emit require.js:1155
    each require.js:57
    emit require.js:1154
EMLEntityView.js:571:25
```

When paused on that exception, the `measurementScale` variable was the string `ratio` rather than an  EML{Non}NumericDomain or EMLDateTimeDomain model, as expected.

laurenwalker:
Ok so I didn't have `enableMeasurementTypeView` configured correctly so now the annotation editor is showing up. But the console error above is still happening, so we'll want to  fix that so drafts  are saved.

amoeba:
Thanks for taking a look @laurenwalker. It looks like the above issue was introduced sometime after I branched off develop. A rebase and a few merges later seemed to fix it. However, I'm seeing a new error `measurementScaleModel.isValid is not a function` which might be related. I had said I was going to get this done by the end of today but I need a tad more time to fix this issue and make sure there isn't a larger issue.

laurenwalker:
Thanks @amoeba for doing that, looks great!

amoeba:
Hey @laurenwalker, I found and fixed a bug I must've introduced by accident in https://github.com/NCEAS/metacatui/pull/1660/commits/65f1a8071a0decb98555b952d79f27e5b28141ec and things are working well for me now. Thanks for the review, feel free to merge away.

Issue #1653, opened 2021-01-26
Semantic search in portals


amoeba:
Portals collection queries can currently be built with filters based on our `sem_annotation` field and portal data searches can search for datasets based upon the field but it's currently limited to manually entering URIs in both cases. This means only experts in our system can really use the feature at the moment.

A much more powerful and user-friendly way to expose semantic search in both (1) portal data search and (2) portal collection query building would be to help the user to find the right term(s) by more advanced means such choosing a value (or values) in a tree browser and/or free-text search against a set of terms, an ontology, or a set of ontologies.

I think we need to discuss a bit what type of scope we might target here before we do any work and follow that with some mock-ups.

- [x] Enhance semantic search in custom portal data search filters
- [x] Enhance semantic search in portal query builder

Edit 20200126: @robyngit informed me that this has been hooked into the query builder and shows up when a BioPortal API key is set. So we might just need to hook this into the portal data search.

robyngit:
Hey @amoeba, if I understand what you're proposing correctly, this feature is already implemented in the Query Builder. However, it only works on the ADC portal builder because it requires a bio-portal API key to be configured (`MetacatUI.AppConfig.bioportalAPIKey`). If one is not configured, like on the KNB or DataONE MetacatUI, then it reverts back to the simple text input where you would need to manually enter a URI, like you pointed out.

> choosing a value (or values) in a tree browser

<img width="430" alt="Screen Shot 2021-01-26 at 09 24 54" src="https://user-images.githubusercontent.com/26600641/105857983-d78e9900-5fb8-11eb-8401-53bdf770528e.png">

> free-text search against a set of terms

<img width="430" alt="Screen Shot 2021-01-26 at 09 25 16" src="https://user-images.githubusercontent.com/26600641/105857999-da898980-5fb8-11eb-9665-a497d0568806.png">

Is this what you meant, and is it showing up for you when you build a query through the ADC? Were you imagining other ways that we could improve this?

I don't think we're using this view for the custom portal search filters yet

amoeba:
ah cool, thanks @robyngit. I had missed that as I was only testing on test.arcticdata. That's totally what we were thinking. I'll update the issue text to reflect this. We would like to also see it exposed as a custom portal search filter so I'll leave this open to track that work.

rushirajnenuji:
We demoed this feature with the Salmantics team, here are some of the enhancements suggested:

[...]

PR #1842, opened 2021-07-29
Feature #1653 portal semantic search

rushirajnenuji:
Adds the ability to have a custom semantic annotation search filter for portals (as defined in the portal schema). Displays a dropdown that presents an annotation tree view for the portal and includes ability to filter datasets based on the selected inputs.


laurenwalker:
I notice that @amoeba has been tagged as the reviewer. @rushirajnenuji , are you waiting on Bryce's review and/or my review for general metacatui release?

Does this work depend on any new development in Metacat (such as the annotation indexing changes in [NCEAS/metacat#1517](https://github.com/NCEAS/metacat/issues/1517))?

amoeba:
Heya @laurenwalker, I was hoping to have a first look here just for the basics and then thought we could run it by you and @robyngit for a review of it relative to portals in general. How does that sound? If you do want to have a look now, feel free otherwise I can ping you when I'm done.

laurenwalker:
Ok, sounds good, thanks

amoeba:
Hey @rushirajnenuji, I tested this out and it seems to work pretty well. I did notice a few things that were a bit inconsistent:

- The placeholder value from the portal document isn't used
- The icon specified in the portal document isn't used
- The overall look and feel doesn't match other filters
  Screen Shot 2021-08-04 at 12 50 17 PM
- Clicking into the filter jumps the datasets list down which is a bit jarring
  
- You can lose track of the reset/up buttons in the tree

I haven't reviewed the code just yet but I think these visual things should be addressed. Since you're 100% on KWG, how would you feel if I worked on tackling these and wrapping up this PR?

rushirajnenuji:
Thank you @amoeba  for reviewing these changes. I can work on them this week and update the PR or wrap the PR.

rushirajnenuji:
Hi @amoeba,
update: I've pushed few commits addressing the above suggestions. I'm still seeing issues with the datasets being pushed down. I'll add that fix asap.

rushirajnenuji:
Hi @amoeba  - I've pushed in the changes for the above suggestions. I'm still thinking of ideas for `You can lose track of the reset/up buttons in the tree` on how to improve the experience. It would be great if you could please review the changes whenever you can spare some time. Thank you!

amoeba:
Great, thanks @rushirajnenuji. I'll have a look.

amoeba:
This works great @rushirajnenuji, nice work. I noticed a few display issues that should be quick to fix so I'll just make those on my end shortly and try to wrap this up. I haven't looked at the implementation side of this just yet so I'll be sure to do that and drop comments here too.

src/js/views/filters/FilterGroupView.js
```diff
@@ -72,6 +75,47 @@ define(['jquery', 'underscore', 'backbone',
...
+            var newValues = _.clone(filter.get("values"));
+            // append to values array if not exist
+            if (newValues.indexOf(term) == -1) {
+              newValues.push(term);
+              filter.set({"values": newValues});
+            }
+
+            view.trigger("addNewAnnotationSearch", event, item, filter);
+          });
+        }
```
amoeba, line 118: This works fine but I thought I'd point out that this logic could be encapsulated within a AnnotationFilter model to make it line up better with the rest of the logic so then it could just be `var filterView = new AnnotationFilterView({ model: filter });`. Part of why this is the way it is is because the AnnotationFilterView lives in pre-Portal-era times so it's serving multiple roles.

@laurenwalker let me know if you feel strongly about this

amoeba (commented):
I made some tweaks to make the styling of the annotation filter visually consistent with the other filters and also to fix some display issues. I've tested this pretty thoroughly and it appears to be working just fine.

There's a bit more annotation-specific code in here that might be more generalize-able with some refactoring but this might just be good enough for now.

This is good enough to pass over to @laurenwalker for a second review. @laurenwalker remember to set a bioportal API and feel free to test with my `bryceannotationtest` portal on `test.arctic`. I had to skill up on how the portal and filters code worked so I'd appreciate a set of eyes on that.

src/css/metacatui-common.css
```diff
@@ -6278,6 +6293,41 @@ body.mapMode{
...
+}
+
+.filter.annotation-filter .ui.multiple.search.dropdown > input.search {
+	font-size: inherit;
+	line-height: 1em;
+	min-height: 1em;
+	margin: 0.4em;
+}
+/* End Overrides for default Bootstrap styling for the searchable select */
+
```
amoeba, line 6330: Just a note to look this over, @laurenwalker. There's a matching commit that explains more of the rationale too. I opted for this approach instead of refactoring the SearchableSelect styling.

laurenwalker:
I agree with Robyn about the annotation fields. I imagine as we index more and different annotations, we will want to support Filters for them without needing to recode this.

laurenwalker:
I tested it out and it looks good to me. 

One thing I would suggest changing is limiting the filter type options to only the Free Text option. Otherwise it may be confusing to users who are building the filter and may not know what to put for the dropdown options in the dropdown filter type, for example.



I think ideally we would want to show a preview of the annotation filter there instead, but for now I think showing the default Free Text is fine.

Robyn did some work recently where some of these options are disabled depending on the field type, so maybe she can point out how to implement this.

robyngit:
> Robyn did some work recently where some of these options are disabled depending on the field type, so maybe she can point out how to implement this.

The UI types that a user can select from are configured in `FilterEditorView.uiBuilderOptions`:

https://github.com/NCEAS/metacatui/blob/da1b6d188700bf38ec06a08be430beafad252d99/src/js/views/filters/FilterEditorView.js#L250-L315

Here is the code that uses that config to allow/block each UI based on the **type** of field(s) selected:

https://github.com/NCEAS/metacatui/blob/da1b6d188700bf38ec06a08be430beafad252d99/src/js/views/filters/FilterEditorView.js#L703-L713

Right now, the `FilterEditorView` only limits filter UIs based on [type](https://github.com/NCEAS/metacatui/blob/da1b6d188700bf38ec06a08be430beafad252d99/src/js/models/queryFields/QueryField.js#L261-L268) ([see also](https://github.com/NCEAS/metacatui/blob/7abdff27c801cbbddd70f76bf4eacb90506d7628/src/js/collections/queryFields/QueryFields.js#L123)). The view will need to be updated to limit based on the solr field name as well, similar to [how the QueryRuleView works](https://github.com/NCEAS/metacatui/blob/main/src/js/views/queryBuilder/QueryRuleView.js#L396).

amoeba:
I've knocked out a few of the smaller things here and search is working but editing isn't yet. We hadn't initially scoped being able to enable/edit semantic search through the UI as part of this work but it really looks like we need to to make this all work well. 

I had a look at how all of that works and I think we need a fully Portal-aware Filter and FilterView model here which we don't have yet. So I'm going to work on those.

laurenwalker:
Since there is still more work to be done, I am going to close this PR and we can reopen it when it is ready for review. That will be much simpler for me so I know when it's ready to merge. Updates can go in the individual Issues.

amoeba:
Heya @laurenwalker and @robyngit, this is ready for another review whenever you have time. Since you both last reviewed, I refactored things a fair bit, mainly to make this code follow the patterns you two have already laid out with the Filter* models and views. I've put changes in granular commits that include explanations I hope you'll find useful. The key things are:

- I fixed some remaining styling issues so instances of the AnnotationFilterView should look correct everywhere where they didn't early in this PR
- AnnotationFilterView now supports customization of the ontology and startingRoot and portal documents can make use of this with `<option>`s
- Semantic search in portals now uses a new SemanticFilterView which lives in `views/filters` and is my attempt at laying the groundwork for having more types of semantic filters UIs beyond the BioPortal tree widget-based view that's implemented in AnnotationFilterView. ie SemanticFilterView could render more than just AnnotationFilterValue (our tree widget) in the future, based upon portal config
- I've added a mechanism to restrict which portal UI builders can be used with a set of fields and configured it so only the Free text builder shows when the user has selected sem_annotation
- The Filter model has a new attribute, `valueLabels`, which lets filters set alternative or human-readable labels for `values`. Semantic search uses this to show rdfs:labels instead of URIs in the UI

To test, I've set up a more comprehensive Portal document on test.arcticdata.io which you can check out with a config pointing at test.arctic and your bioportalAPIKey set:

```
{
  ...
  baseUrl: "https://test.arcticdata.io",
  d1CNBaseUrl: "https://cn-stage.test.dataone.org",
  bioportalAPIKey: "$YOUR_KEY",
  ...
}
```

And then run the dev server and visit http://localhost:3000/portals/portal-semantic-search-playground. 

See the toggle below for the manual test scenarios I ran through. I think it's fairly comprehensive.

<details>
<summary>Test scenarios</summary>

ANNOTATION FILTER

- [x] Should default to ECSO#MeasurementType
- [x] Should support another ontology with no starting root
- [x] Should support another ontology with a startingn root
- [x] All of these changes shouldn't break the normal search catalog annotation
      search

PORTAL SEARCH

- [x] Repeated side-by-side semantic search widgets should look right
- [x] Semantic search should look right as the $POS filter where:
  - [x] First
  - [x] Middle
  - [x] Last
  - [x] Single
- [x] A semantic search filter should respect label and placeholder values
- [x] A semantic search filter should use fallback values for label and
placeholder so there's always something there
- [x] When the user picks a term, the dropdown should hide and any text should
      be cleared

PORTAL UI BUILDER

- [x] Three side-by-side semantic search widgets should look right
- [x] Semantic search should look right as the $POS filter where:
  - First
  - Middle
  - Last
  - Single
- [x] A semantic search filter should be editable
- [x] A semantic search filter should be removeable
- [x] A semantic search filter should be addable
- [x] Only the Free text builder optino should be available if any field is
      sem_annotation

COLLECTION QUERY BUILDER

- [x] A user should be able to add a collection query for a single term
- [x] A user should be able to add a collection query for multiple terms
  - [x] Terms can be AND'd
  - [x] Terms can be OR'd
- [x] If a user adds a non-semantic field to the same rule, the UI should switch
  back to free-text and retain the value
- [x] If a user has a semantic a non-semantic field in a single rule, removing the
  non-semantic field should return the filter back to a semantic filter

</details>

src/js/models/AppModel.js
```diff
@@ -1644,7 +1653,14 @@ define(['jquery', 'underscore', 'backbone'],
               placeholder: "Geographic region",
               icon: "globe",
               description: "The geographic region or study site, as described by the submitter"
-            }
+            },
+            {
```
laurenwalker, line 1658: The inline documentation for this config needs to be updated. This is no longer used by the Edit Collection View ever since we updated it to use the new Query Builder. This config will be used in the future when we replace the old DataCatalogView with the DataCatalogWithFilters view. I don't think we want an AnnotationFilter to be a default filter for that view since not all repos will ahve a bioportal API key enabled. So let's remove this here

laurenwalker (changes requested):
Hi @amoeba - Thanks for the thorough write-up for your PR, it was very helpful for my review. Everything is looking great so far. I experienced only a couple issues during my testing, see below. Check marks are next to the tests that passed for me, unchecked boxes had issues which are described inline:

ANNOTATION FILTER

- [X] Should default to ECSO#MeasurementType
- [X] Should support another ontology with no starting root
- [X] Should support another ontology with a starting root
- [X] All of these changes shouldn't break the normal search catalog annotation
    search

PORTAL SEARCH

- [X] Repeated side-by-side semantic search widgets should look right
- [x] Semantic search should look right as the $POS filter where:

    - [X] First
    - [X] Middle
    - [X] Last
    - [x] Single

- [X] A semantic search filter should respect label and placeholder values
- [X] A semantic search filter should use fallback values for label and placeholder so there's always something there
- [X] When the user picks a term, the dropdown should hide and any text should be cleared

PORTAL UI BUILDER

- [x] Three side-by-side semantic search widgets should look right
- [x] Semantic search should look right as the $POS filter where:
    - [x] First
    - [x] Middle
    - [x] Last
    - [x] Single
- [x] A semantic search filter should be editable
- [x] A semantic search filter should be removeable
- [x] A semantic search filter should be addable
- [x] Only the Free text builder optino should be available if any field is sem_annotation
- [x] If a user adds a non-semantic field to the same rule, the UI should switch back to free-text and retain the value
- [x] If a user has a semantic and non-semantic field in a single rule, removing the non-semantic field should return the filter back to a semantic filter

COLLECTION QUERY BUILDER

- [x] A user should be able to add a collection query for a single term
- [ ] A user should be able to add a collection query for multiple terms
    - [x] Terms can be OR'd
    - [ ] Terms can be AND'd
        - **Error:** When I create a semantic annotation Rule with two values, I can click the "OR" text and the UI changes to "AND", but it seems like the model never updates because the "Save" button doesn't pop up at the bottom of the Editor, and when I force a save (by editing something else and clicking Save), the resulting Solr query is still using "OR". Other Rules with "AND" work as expected.
        I saved a copy of your portal with those Rules for you to look at (also on test.arcticdata.io): http://localhost:3000/portals/portal-semantic-search-playground-lauren2

Screen Shot 2022-03-16 at 2 35 55 PM

## Other issue:
- One minor UI issue is that in the Collection Query builder, when a semantic annotation Rule is added, the "Type a value" input is a few pixels higher than it should be:

Screen Shot 2022-03-16 at 2 27 46 PM

src/js/views/filters/FilterEditorView.js
```diff
@@ -270,6 +273,7 @@ define(['jquery', 'underscore', 'backbone',
             iconFileName: "choice.svg",
             description: "Allow people to select a search term from a list of options",
             filterTypes: ["filter"],
+            blockedFields: ["sem_annotation"],
```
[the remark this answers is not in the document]
amoeba: Great idea. I made the change two all three instances of this in 068b0ed596b0508320051e316b2fa9bc093075e7.

src/js/models/AppModel.js
amoeba: Sounds good. Done in 1987564b73681bc2ca45743e4193aa07a54683d0. I tested that this doesn't break semantic search in the DataCatalogView and it looks fine.

amoeba:
Thanks for going through this for us @laurenwalker. I addressed your two review questions in 1987564b73681bc2ca45743e4193aa07a54683d0 and 068b0ed596b0508320051e316b2fa9bc093075e7.

For the Other issue above, I fixed it in b3bac9cac322faeef0899b5029f105ba0d47fa96 to make it look like this:

Screen Shot 2022-03-16 at 5 13 56 PM

That still leaves us with a minor mis-alignment issue but it actually seems to affect all query rules, not just semantic annotation rules. I assume you were talking about the more major of the two?

The only thing remaining here is that I'm still looking at the AND/OR bug you caught and, while doing so, I found another bug: For a new portal with on filter rules, if you add just a semantic filter, the datasets list below doesn't automatically update like it does for other fields. I'm guessing this is all something pretty fundamental with the way we're setting things up for semantic filters. I'll try to get this back to you pretty quick here. Thanks again!

amoeba:
Thanks for the offer @laurenwalker. I took a look at it and I'm a bit stumped and, if you have time, another set of eyes would be great. It's pretty clearly an issue with how the `AnnotationFilterView` is working because swapping it out for a plain-old `SearchableSelectView` makes things work fine. I took some notes:

- The biggest lead I have is that it's just your first selection that doesn't update the results. If you remove your first selection or add another, things update. Oddly, the `separatorChanged` event doesn't also start working.
- To get the `separatorChanged` event from the `SearchableSelectView` to fire all the way up to QueryRuleView, I had to manually re-trigger it, e.g.,

    ```
    # In AnnotationFilterView#createMultiSelect
    view.multiSelectView.on("separatorChanged", function(newVal) {
      view.trigger("separatorChanged", newVal);
    });
    ```

    This doesn't match my mental model for how event propagation should work.
- I'm getting a double firing of `QueryRuleView.handleValueChange` when choosing an item in a semantic annotation query rule. I don't get a double fire for the `separatorChanged` event which is odd.

Do you have a sec to take a look @laurenwalker ?

amoeba:
One small usability thing @jeanetteclark just found is that you can't just paste in a URI and select that as the value. It tries to find the term in the ontology picker. I'm not sure if we want to support this but it'd be nice to.

laurenwalker:
I've fixed the two aforementioned outstanding bugs:
> [The only thing remaining here is that I'm still looking at the AND/OR bug you caught 
](https://github.com/NCEAS/metacatui/pull/1842#pullrequestreview-912175114)
and

>[ your first selection that doesn't update the results.](https://github.com/NCEAS/metacatui/pull/1842#issuecomment-1077015339)

which are fixed in the two commits referenced in this Github thread. I will create an issue for this particular functionality:

> One small usability thing @jeanetteclark just found is that you can't just paste in a URI and select that as the value. It tries to find the term in the ontology picker. I'm not sure if we want to support this but it'd be nice to.

which I think can wait until an iterative release.

amoeba:
Thanks a ton @laurenwalker. I'm not sure I woulda figured out 517208dea5ec19dfc180e7bd652e3eefa2a67641.

PR #1843, opened 2021-08-03
Refactor AnnotationView to show two separate popovers instead of a single one

amoeba:
Before this change, clicking anywhere in the annotation pill shows a single popover providing more information about the `valueURI`. No extra information can be found about the `propertyURI` beyond its label.

Screen Shot 2021-08-02 at 5 56 02 PM

Now, you get two separate popovers, and each one can show more information about each term (if found):

Screen Shot 2021-08-02 at 5 57 33 PM

This change is tied with https://github.com/NCEAS/metacat/pull/1518 which modifies the underlying DOM structure returned by the view service. For perfect functionality, this change requires a Metacat upgrade. However, the changes were made to be relatively backwards compatible:

| | Old Metacat | New Metacat |
|-:|:-:|:-:|
|**Old MetacatUI** | ✅ | Pill has one minor glitch, old popover works |
|**New MetacatUI** | Pill looks good, no popovers | ✅ |

The minor glitch above when an old MetacatUI encounters an upgraded Metacat is just that the right side of the pill doesn't have a rounded border. 

**Testing notes**

1. I don't have an example of a real dataset that exercises the new behavior so I put a test document on my own VM with a non-sense annotation just to show everything works: https://neutral-cat.nceas.ucsb.edu/view/a534573b-c40b-47ad-909f-d7a639e6a003

2. Ensure `bioportalAPIKey` is set in your config to a valid BioPortal API key

    Here's my usual config:

    ```json
    MetacatUI.AppConfig = {
      root: "/",
      theme: "arctic",
      baseUrl: "https://neutral-cat.nceas.ucsb.edu",
      d1CNBaseUrl: "https://cn-stage.test.dataone.org",
      bioportalAPIKey: "CHANGEME",
      metacatContext: "metacat",
    }
    ```
3. https://neutral-cat.nceas.ucsb.edu has the latest changes both to Metacat and to MetacatUI deployed on it and is in cn-stage and you're welcome to test things there

### DIFF
--- a/src/css/metacatui-common.css
+++ b/src/css/metacatui-common.css
@@ -2739,8 +2739,7 @@ section#user-citations {
 }
 /** Annotation pills **/
 .annotation .annotation-property,
-.annotation .annotation-value,
-.annotation .annotation-findmore {
+.annotation .annotation-value {
   padding: 3px 6px;
 }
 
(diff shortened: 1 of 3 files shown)


laurenwalker:
Excellent write-up Bryce, thank you! I'll get merged once NCEAS/metacat#1518 is merged.

laurenwalker:
Just tested this and it works great for me. The only change I'd suggest we make is moving the popovers so that they are not overlapping, so they're both readable at the same time. (#1874) But I don't think that should hold up a first release. I'll merge and close this PR once Metacat 2.15.1 is closer  to release.

amoeba:
Sounds great, thanks for taking a look @laurenwalker.

PR #1956, opened 2022-02-25
Fix URL encoding issues for GET requests for Solr searches

kitsunenokenja:
Most of the main menu selections in the UI were afflicted by bad GET
request composition including characters not permitted by the HTTP
standard, which was responsible for the server rejecting the request
with HTTP 400.

These patches resolve the bad request failures by dropping the invalid
injection of backslashes, and substituting special characters with the
URL-encoded representations, primarily square brackets. Lastly, there
was an instance where a Solr query fragment was being composed with
OrcID strings that weren't enclosed in quotes, which is also fixed.

The late string manipulations have been marked as TODO items since a
more permanent, ideal fix is to rework the preceding logic rather than
stripping bad characters back out of a string that was just composed.

Jing and Lauren are aware that we have been working on this issue. After
overcoming the problems causing the error 400 responses, we're receiving
good responses back, indicating that additional character escaping for Solr's
input was not even necessary. For that reason, I completely removed the
`escapeSpecialChars` methods.

I don't know the reason behind not simply making all the complicated requests
POSTs, regardless of their size or length; it would simplify a lot of the code.

### DIFF
--- a/src/js/models/Search.js
+++ b/src/js/models/Search.js
@@ -354,10 +354,10 @@ define(["jquery", "underscore", "backbone", "models/SolrResult", "collections/Fi
                       }
 
                       if( forPOST ){
-                        query += this.fieldNameMap["id"] + ':*' + this.escapeSpecialChar(identifiers) + "*";
+                        query += this.fieldNameMap["id"] + ':*' + identifiers + "*";
                       }
                       else{
-                        query += this.fieldNameMap["id"] + ':*' + this.escapeSpecialChar(encodeURIComponent(identifiers)) + "*";
+                        query += this.fieldNameMap["id"] + ':*' + encodeURIComponent(identifiers) + "*";
                       }
                     }
                 }
@@ -493,7 +493,7 @@ define(["jquery", "underscore", "backbone", "models/SolrResult", "collections/Fi
                             " AND endDate:[* TO " + yearMax + "-12-31T00:00:00Z]";
                     }
                 }
-                
+
                 //----- public/private ------
                 if (this.filterIsAvailable("isPrivate") && ((filter == "isPrivate") || getAll)) {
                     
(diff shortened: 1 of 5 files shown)


laurenwalker:
Thanks for the PR. I reviewed your changes and made a [couple](https://github.com/NCEAS/metacatui/commit/f3261d00c2ed80ea47bea09cee98f8e57e4765f4) of [commits](https://github.com/NCEAS/metacatui/commit/946a4c5bfbf78a77e4b39a2358561d1a5fa757b8) that incorporate your suggestions but with changes.

We still need to [escape reserved Solr characters](https://solr.apache.org/guide/7_3/the-standard-query-parser.html#escaping-special-characters) in search terms, so we cannot simply remove that method. I changed it so that it escapes with URL-encoded backslashes, which should fix the issue where Tomcat responds with an error.

I incorporated your change to wrap ORCIDs in quotes, but expanded that to [check for DataONE groups ](https://github.com/NCEAS/metacatui/commit/f3261d00c2ed80ea47bea09cee98f8e57e4765f4#diff-7e7534544974514571e423f056d9c67c40c5def488eae082e4faccebbab80787R475)as well.

@kitsunenokenja I tested these changes with MetacatUI deployed as a Tomcat webapp, but it would also be nice if you could verify that the `develop` branch works in your environment now.

----

While testing my changes, I found that Metacat must strip Solr special characters (escaped or not) from queries, which is why I think queries still work in your MetacatUI fork where you removed `escapeSpecialCharacters()`. I created a [Metacat issue](https://github.com/NCEAS/metacat/issues/1575) to address that. I also found a bug where [ampersands in query strings cause errors](https://github.com/NCEAS/metacat/issues/1576), even though they should be accepted if they are properly escaped with backslashes. MetacatUI will now remove ampersands from query strings until that bug is resolved.

> I don't know the reason behind not simply making all the complicated requests
POSTs, regardless of their size or length; it would simplify a lot of the code.

I agree that we should just send all queries as POSTs. The reason why we initially chose to support GETs is because POSTs to the Solr service weren't supported until Metacat 2.10.2 and we wanted to allow MetacatUI owners to upgrade the UI witout requiring a upgrade to Metacat right away. Given that was 3 years ago, I think it's time to [drop support for GETs to Solr](https://github.com/NCEAS/metacatui/issues/1998) via MetacatUI.

PR #1959, opened 2022-03-09
Update the urls.packageServiceUrl to bagit 1.0

ThomasThelen:
A small one line change to bump the bagit version to 1.0


laurenwalker (changes requested):
@ThomasThelen Let's make the bagit format a configurable option in the AppConfig/AppModel. That way the MetacatUI release will not depend on a Metacat upgrade yet.

ThomasThelen:
> Quick thought: This might be better named as `packageFormat` as the DataONE `getPackage` service anticipates other non-bagit package formats in the future.

Ah you're totally right good catch! I'll rename it

PR #1981, opened 2022-03-18
Enable layer-level notifications in Cesium map

### LINKED ISSUE #1972: Enable adding a notification to individual layers in the Cesium map
Need the ability to indicate that layers are in development. The notification could also be used to draw attention to a new layer, or to indicate that a layer has been recently updated.

robyngit:
This PR adds support for configuring notifications in Cesium maps that are specific to individual layers. A notification is configured by adding a `notification` property to a layer, for example:

``` json
{
  "layers": [
    {
      "label": "Ice Wedge Polygons",
      "type": "Cesium3DTileset",
      "notification": {
        "badge": "Preview",
        "style": "green",
        "message": "Pan-arctic coverage will be available for this layer soon! While the data pipeline is under development, a subset of the data is available here."
      }
    }
]}
```

This adds a badge element next to the layer name in the layers list, and a longer message in the layer details panel:

Screen Shot 2022-03-18 at 13 43 29
Screen Shot 2022-03-18 at 13 43 07

Five styles of notifications are available:

Screen Shot 2022-03-18 at 13 29 34

The image above comes from the the test portal `permafrost-test-portal` on `urn:node:mnTestARCTIC` - this portal could be used to test the notifications feature.

If we move forward with these notification styles, we should test out how they look with the new "Zoom To" button styles, if PR #1976 is accepted.

This PR additionally refines spacing & font sizes in the UI elements of the Cesium Map. It became apparent when adding the badge element that the UI layout could be a little more compact (I think elements were too large and spaced out before.)

Fixes #1972

### DIFF
--- a/src/css/map-view.css
+++ b/src/css/map-view.css
@@ -33,6 +33,9 @@
   --map-col-buttons: #313c52;
   --map-col-text: #F9FAFB;
   --map-col-highlight: var(--portal-primary-color);
+  --map-col-green: #069868;
+  --map-col-blue: #057fc6;
+  --map-col-yellow: #f8b406;
   /* SIZING: */
   --map-size-toolbar-link: 2.2rem;
   --map-size-toolbar-link-margin: 0.4rem;
@@ -44,6 +47,13 @@
   --map-shadow-md: 0 1px 9px -1px rgba(0, 0, 0, 0.2), 0 1px 2px 0px rgba(0, 0, 0, 0.5);
   /* NOTE: 768px is used as the mobile -> desktop breakpoint throughout the map view, but we cannot use a CSS variable for this. */
 }
+
+/* hide the credits until we can find a better placement for them */
+.cesium-widget-credits, .cesium-credit-lightbox-overlay {
+  display: none !important;
+}
+
+/* ---- TOOLTIP ---- */
 
 .map-tooltip {
   color: var(--map-col-text);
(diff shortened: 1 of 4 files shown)


laurenwalker:
@robyngit I merged your changes to develop, could you test that my merge was ok given this:

> If we move forward with these notification styles, we should test out how they look with the new "Zoom To" button styles, if PR https://github.com/NCEAS/metacatui/pull/1976 is accepted.

If everything looks ok I'd like to tag a release with this tomorrow.

Thanks!

robyngit:
@laurenwalker - I fixed a little CSS issue with the merge, and tested with the two new features together. I think it looks okay, though a second opinion would be great! The blue zoom button combined with the blue notification style is not great. This happens because the button color comes from the `portal-secondary-color`, which, in the case of the PDG portal, happens to closely matches the blue notification color. I think it will be somewhat up to portal creators to make good design decisions (e.g. avoid the blue notification style if you chose a similar blue for the secondary portal color.) 

## Default
Screen Shot 2022-03-22 at 13 28 52

## Contrast
Screen Shot 2022-03-22 at 13 29 36

## Blue
Screen Shot 2022-03-22 at 13 29 30

## Green
Screen Shot 2022-03-22 at 13 29 24

## Yellow
Screen Shot 2022-03-22 at 13 29 18

laurenwalker:
@robyngit I agree that the two blue boxes next to each other make them both look like buttons (or neither look like buttons). Out of all the options, I like the green notification box for our "Preview" notification.

Issue #1955, opened 2022-02-21
Latitude and Longitude query values only accept integers


laurenwalker:
Query rules with latitude and longitude fields are only accepting integers, but they should accept floats. This is preventing portal users from narrowing their search down to exact geographic areas. 

![Screen Shot 2022-02-21 at 2.10.21 PM.png](https://images.zenhubusercontent.com/5d152e12c27d0c306f021863/b2f58029-c404-494d-b6da-739e5530bc56)

PR #2003, opened 2022-04-08
Enable the NumericFilter to handle coordinate queries

robyngit:
Not only should query rules with latitude and longitude fields accept floats, as pointed out by issue #1955, but they should also have a `minRange` and `maxRange` set to -90 to 90 for latitude fields, and -180 to 180 for longitude fields. This PR fixes both issues.

**Summary:**
- Coordinate fields are defined by `queryLatitudeFields` and `queryLongitudeFields` in the `appModel`
- When a `NumericFilter` is initialized with only coordinate fields, or its fields are updated to include only coordinate fields, then:
  - the `rangeMin` is set to -90 or -180
  - the `rangeMax` is set to 90 or 180; and
  - the `step` is set to 0.00001
  - `min` and `max` values that are `< rangeMin` or `> rangeMax` are set to `rangeMin` or `rangeMax`
- The above only happens if each attribute is initially the default for a NumericFilter model (e.g. `rangeMax` will only be updated to 90 when it's fields switch to only latitude fields IF `rangeMax` is `null` (the default value) to start)
- Similarly, when a `NumericFilter`'s fields change, and it no longer contains only coordinate fields, then the `rangeMin`, `rangeMax`, and `step` are set back to the regular defaults for a numericFilter. The `min` and `max` are also rounded to the same number of decimal places as `step`
- The `NumericFilterView` is re-rendered when these changes happen in the `NumericFilter` model, such that a user may now user the slider to set coordinate queries, and may set float values:
  Screen Shot 2022-04-08 at 15 18 56
- Also added unit tests for the `NumericFilter` model, especially tests that pertain to the new features described above

@laurenwalker, I made a decent number of changes to accomplish this, so I was hoping you might be able to give this strategy a quick look over if you have time!


laurenwalker (approved):
Looks good to me, Robyn. I found the step of `0.00001` to be a little tedious to use for the number inputs, but it is working fine for me. The test looks good too, thanks!

PR #2022, opened 2022-06-03
Feature #2021 imagery bounds

### LINKED ISSUE #2021: Support defining a view rectangle for imagery layers
If a view rectangle is not supplied in the [`ConstructorOptions`](https://cesium.com/learn/cesiumjs/ref-doc/WebMapTileServiceImageryProvider.html#.ConstructorOptions) for an imagery layer in Cesium, then Cesium assumes that the layer covers the entire globe. This means that for smaller imagery layers, the "Zoom To" button won't work. We need to allow configuration of the bounding coordinates for imagery layers, and convert these to an instance of the `Cesium.Rectangle` class to pass when constructing the imagery layer.

robyngit:
This PR fixes #2021, allowing us to define the bounding box around an imagery layer when configuring a Cesium Map view.

@laurenwalker, the changes I made to the Cesium models are pretty straight forward, but I was wondering if you could let me know whether the testing strategy I used for imagery models is okay. I uploaded some very small web-tiles to a new `test/data/` directory to use with the test `CesiumImagery` model. The data follows the same directory structure as the models and tests (the data is in sub-directories `models/maps/assets/CesiumImagery`). Is this a good way to include data for the unit tests, should I move it somewhere else, or should I use some external data sources?


laurenwalker:
Thanks Robyn, that test looks good to me and having some sample data for unit tests like the way you have done it is perfect.

Issue #2024, opened 2022-06-08
Refactor UserView with group management separated into a new view


vchendrix:
In order to support Project Spaces team management,  it is necessary to refactor MetacatUI [`views/UserView`](https://github.com/NCEAS/metacatui/blob/main/src/js/views/UserView.js) to separated out the group management functionality so that it can be used outside of the user settings page.

This will require us to make copy of [`views/UserView`](https://github.com/NCEAS/metacatui/blob/main/src/js/views/UserView.js) and then create a subview  `js/views/UserGroupView` that  encapsulates all of the group logic from the UserView.

**Proposed New ESS-DIVE Theme Views**
+ `src/js/views/UserView` (refactored)
+ `src/js/views/UserGroupView` (new)

The outcome of this will not have any functionality changes it will merely replace the MetacatUI view with the new ESS-DIVE theme views above. Additionally, try to change as little logic as possible except where the GroupView interacts with the UserView.

vchendrix:
In order to support Project Spaces team management,  it is necessary to refactor MetacatUI [`views/UserView`](https://github.com/NCEAS/metacatui/blob/main/src/js/views/UserView.js) to separated out the group management functionality so that it can be used outside of the user settings page.

This will require us to make copy of [`views/UserView`](https://github.com/NCEAS/metacatui/blob/main/src/js/views/UserView.js) and then create a subview  `themes/essdive/views/UserGroupView` that  encapsulates all of the group logic from the UserView.

**Proposed New ESS-DIVE Theme Views**
+ `themes/essdive/views/UserView` 
+ `themes/essdive/views/UserGroupView` 

The outcome of this will not have any functionality changes it will merely replace the MetacatUI view with the new ESS-DIVE theme views above. Additionally, try to change as little logic as possible except where the GroupView interacts with the UserView.

laurenwalker:
This would be really useful for other repositories as well, so if this refactor and new view can be done in the main MetacatUI codebase, we would certainly take advantage of it.

laurenwalker:
@vchendrix No, it's something that's on my "Wish List" but I haven't started any development yet.

vchendrix:
@laurenwalker thanks for the clarification.  Should @mamelara branch off main?

laurenwalker:
Yes, main would be best to branch from. Thanks!

laurenwalker:
[...]

PR #2032, opened 2022-06-29
Adds Project Base Model, View and Placeholder HTML

### LINKED ISSUE #1945: Add Project Backbone Model to access project rest api
**Describe the feature you'd like**
Add Project Backbone Model to access project rest api. 

**Is your feature request related to a problem? Please describe.**
We need to be able to view and manage projects in metacat UI.

**Additional context**
The current project JSON is as follows.  We will definitley need to have a identifier that is resolvable 

```json

{
       "url": "https://mule-stage.ess-dive.lbl.gov/project/1/",
        "id": 1,
        "project_title": "Soil Carbon Biogeochemistry",
        "sponsor_program": "TES",
        "project_type": "Lab Awards",
        "short_name": "",
        "pis": [
            {
                "last_name": "Bailey",
                "first_name": "Vanessa",
                "institution": "Pacific Northwest National Laboratory",
                "email": "Vanessa.Bailey@pnnl.gov"
            }
        ]   
 }
```

helbashandy:
Closes #1945

* Adds a base Project model and collection that can be populated
 either from a default or a direct API. In the case of populating
 from the API, extending member nodes will have to extend the Parse
 function if any data structure handling is needed.

* Adds tests for project model and collection.

* Adds a Project view with some functinality to handle basic view
project selection and rendering.

* Adds a projects info template as a placeholder for extending nodes.

Co-authors: Mario Melara, Val Hendrix

### DIFF
--- a/src/js/collections/ProjectList.js
+++ b/src/js/collections/ProjectList.js
@@ -1,0 +1,71 @@
+/* global define */
+"use strict";
+define(['jquery', 'backbone', 'models/Project'],
+    function ($, Backbone, Project) {
+        'use strict';
+        /**
+         @class ProjectList
+         @classdesc A ProjectList represents a collection of projects. This can be
+         used for a projects list view populating EML projects. It also supports loading
+         projects from a third-party API in case projects information is located outside of
+         metacat.
+         * @classcategory Collections
+         @extends Backbone.Collection
+         @constructor
+         */
+
+        var ProjectList = Backbone.Collection.extend({
+            model: Project,
+            type: "ProjectList",                            //The name of this type of collection
+            authToken: undefined,
+            urlBase: undefined,
+            urlEndpoint: "project/",
+
+            /** Builds the url from the urlBase **/
+            url: function(){
+                if(this.urlBase)
+                {
+                    return new URL(new URL(this.urlBase).pathname + this.urlEndpoint, this.urlBase).href
+                }
+                else {
+                    return undefined
+                }
+            },
+            /**
+             * Override backbone's parse to set the data after the request returns from the server
+             */
+            parse: function (response, options) {
+                // Add any custom data structure code here.
+                return response
+            },
+            /**
+             * Override backbone's sync to set the auth token
+             */
+            sync: function(method, model, options) {
+
+                if (this.authToken) {
+                    if (options.headers === undefined){
+                        options.headers = {}
+                    }
+                    options.headers["Authorization"] = "Bearer " + this.authToken;
+                }
+                if (this.urlBase)
+                    return Backbone.Model.prototype.sync.apply(this, [method, model, options]);
+            },
+
+            /**
+             * Initializing the Model objects project variables.
+             */
+            initialize: function (options) {
+                if (MetacatUI && MetacatUI.appModel) this.urlBase = MetacatUI.appModel.get("projectsApiUrl");
+                if (options) {
+                    if (options.authToken) this.authToken = options.authToken;
+                    if (options.urlBase) this.urlBase = options.urlBase;
+                }
+                Backbone.Collection.prototype.initialize.apply(this, options);
+            }
+        });
+
+        return ProjectList
+
+    });
--- a/src/js/models/AppModel.js
+++ b/src/js/models/AppModel.js
@@ -1450,6 +1450,18 @@ define(['jquery', 'underscore', 'backbone'],
       },
 
       /**
+       * Add an API service URL that retrieves projects data. This is an optional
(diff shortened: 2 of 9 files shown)


helbashandy:
Hi @laurenwalker - From our discussion with @mbjones, we figured that the best timeline that works with our deliverables is that we merge this PR in a minor release within the 2.x.x major before July 30th. The same timeline applies to https://github.com/NCEAS/metacatui/issues/2024 which @mamelara will submit a PR for soon. 

Thanks for taking the time to take a look at the PR and please let me know about any feedback for future contributions.

laurenwalker:
Thanks @helbashandy and @mamelara, I will review these as soon as I can!

laurenwalker (changes requested):
Thanks @helbashandy, this looks great. My only requested change is to add class level documentation for the new Project, ProjectList, and ProjectView classes via JSDoc (see https://github.com/NCEAS/metacatui/blob/main/CONTRIBUTING.md#documentation). 

I've also gradually been trying to use UnderscoreJS less and less in MetacatUI in favor of vanilla JS so we can remove it completely someday. That's not a necessary change for your PR, but would be helpful! (e.g. replacing _.each with a for loop or Array.forEach())

helbashandy:
Thanks for reviewing @laurenwalker - I made the suggested documentation changes and removed UnderscoreJS from the model and collection. 

I left it as is on the view as it seems like I have to install other dependencies to load up the template html into the view `template` attribute. Is there a reference you can point me to that compiles the template into a render-able function that uses Vanilla JS that can be used in the view?

laurenwalker:
@helbashandy I've been using JS [template literals ](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals) in place of Underscore's template engine. This does mean that the JS expressions inside the template need to be kept simple (e.g. no big for loops) but I think that is good practice so that our templates aren't performing lots of logic.

helbashandy:
Hi @laurenwalker - I explored removing the whole underscoreJS template syntax completely from the example code of the `projectInfo.html` file, but realized that I would need to have a template literal engine to process the template literal syntax, as the html file is imported as a text string and would need to be evaluated for the template literal to be processed instead of rendering as text. As it seems like MetacatUI didn't adapt any template literal engines yet (literals were used directly in the JS views and not imported from an HTML file), I only replaced the underscoreJS template functions with the Vanilla HTML.

laurenwalker:
Thanks for making those changes Hesham. Using the Underscore templates for now is ok until we come up with a more formal solution.

However the tests are not running on your branch. When I look at the GitHub Actions results, it says the action passed/completed ok but when you look at the run details, it says there are `FAILS: 0 PASSES:0`. When I checked out your branch and ran the tests via `npm view-tests`, there are Javascript errors in the console:

```
The resource from “http://localhost:3001/src/js/js/specs/unit/models/project/Project.spec.js?v=undefined” was blocked due to MIME type (“text/html”) mismatch (X-Content-Type-Options: nosniff).
[test](http://localhost:3001/test/)
Loading failed for the <script> with source “http://localhost:3001/src/js/js/specs/unit/models/project/Project.spec.js?v=undefined”. [test:1:1](http://localhost:3001/test/)
Uncaught Error: Script error for: js/specs/unit/models/project/Project.spec
http://requirejs.org/docs/errors.html#scripterror
    makeError http://localhost:3001/src/components/require.js:163
    onScriptError http://localhost:3001/src/components/require.js:1666
[require.js:163:17](http://localhost:3001/src/components/require.js)
The resource from “http://localhost:3001/src/js/js/specs/unit/collections/ProjectList.spec.js?v=undefined” was blocked due to MIME type (“text/html”) mismatch (X-Content-Type-Options: nosniff).
[test](http://localhost:3001/test/)
Loading failed for the <script> with source “http://localhost:3001/src/js/js/specs/unit/collections/ProjectList.spec.js?v=undefined”. [test:1:1](http://localhost:3001/test/)
Uncaught Error: Script error for: js/specs/unit/collections/ProjectList.spec
http://requirejs.org/docs/errors.html#scripterror
    makeError http://localhost:3001/src/components/require.js:163
    onScriptError http://localhost:3001/src/components/require.js:1666
[require.js:163:17](http://localhost:3001/src/components/require.js)

​```

I will change the GitHub Action so that it wil fail when the tests fail to run at all.

helbashandy:
Hi @laurenwalker - This is strange, I have the tests running with no issues locally when I open up `index.html` of the tests dir.

Screen Shot 2022-07-21 at 9 12 24 AM

PR #2081, opened 2022-12-16
Feature deprecating package table view 1451

### LINKED ISSUE #1451: Deprecate PackageTableView and replace with DataPackageView
Use the DataPackageView in the MetadataView to display the package instead of the PackageTableView. This will make improvements to display easier to support and maintain since it will be contained in a single view. It also helps keep the UI consistent.

rushirajnenuji:
Replaces the PackageTableView in MetacatUI to utilize the DataPackageView for rendering the dataset display in MetadataView 

Closes #1451


robyngit:
Besides the minor CSS differences that you are working on fixing, you succeeded in getting the table to look good and work in the MetadataView. Excellent!

I see that DataPackageView now contains all the original methods and properties from the to-be-deprecated PackageTableView. In essence, the new DataPackageView uses one set of functions to display the editable version of the table, and a completely separate set of functions to display the non-editable version. The two versions of the view also use different templates and different model/collection types.

two-views-in-one

What are the advantages of this "two-views-in-one" method of implementation? Did you consider instead refactoring some of the functions that are already in the DataPackageView to be able to also render and display the non-editable version of the table? In other words, rather than copying & pasting everything from PackageTableView into DataPackageView, why not use the template, render method, collection, and functions that the DataPackageView already has, and update them to render a non-editable version of the table?

Here are some of the benefits I see to a refactor of the DataPackageView over the current two-views-in-one approach:
- Reduce redundancy between the PackageTableView and the DataPackageView. For example, the styles and layout are very similar between the two views. If we define those CSS rules and the HTML template in just one place, rather than two, then it is easier to make consistent updates to to the table styles in the future.
- Make use of what we've developed for one view in the other view. For example, right now the sort function can only be used on the Metadata view version of the table, but it would be nice if we could use it for the Editor version as well.
- Much easier to implement the rest of the Folders in datasets Epic. In particular, the DataPackageView already makes use of the DataPackage collection. A collection will be needed to represent a hierarchical data package. Also, if the two versions of the views are more closely aligned with one another, then we will only need to implement the hierarchical data package display once. As it is right now, we'd basically need to implement it separately: once for the Metadata version of the view, and again for the Editor version of the view. This would also create a lot more redundancy in the view.
- Consistency with Backbone/MetacatUI conventions. It is a very common practice throughout MetacatUI to instantiate a view, then render it with `newView.render()`. Using `render` for the editor version and `renderMetadataView` for the metadata version is unexpected and could make it more challenging for developers to understand how to use this view in the future. A more common practice would be to use `new DataPackageView({ edit: false }).render()` for the metadata version and `new DataPackageView({ edit: true }).render()` for the editor version.

mbjones:
Great comments, Robyn, and I think your reasons for a single view  make a lot of sense.

robyngit:
@rushirajnenuji what is the status of this PR?

rushirajnenuji:
Hey @robyngit  -- from what I remember I _think_ we discussed during one of our dev calls the option of closing this PR and addressing the feature as a whole in the [latest PR](https://github.com/NCEAS/metacatui/pull/2209) (still in draft status). 

Does that still work for you?

Issue #2041, opened 2022-07-26
Multiple portal metric bugs are persisting


mburrus:
**Describe the bug**
Some portals experience no metric issues and all graphs/tables/wheels function as expected. Other portal metric pages have one or more of the following issues. These issues appear to be unique to portal metrics only.
 
I couldn't identify why only some portals have these issues or what is causing just one issue to appear versus all three issues.

1. The download/view/citation tables incorrectly report that the count is 0, rather than stating that the tables are failing to render ([metrics-service#85](https://www.google.com/url?q=https://github.com/DataONEorg/metrics-service/issues/85&sa=D&source=calendar&ust=1659305690349852&usg=AOvVaw0vCW9GufuMf18zlgl944U5), [ess-dive-catalog#455](https://www.google.com/url?q=https://github.com/ess-dive/ess-dive-catalog/issues/455%23issuecomment-888538631&sa=D&source=calendar&ust=1659305690349852&usg=AOvVaw0oX4vrPnfmqIBauMbQRG8G))
2. Some metric graphs don't render (it's inconsistent which graphs don't work) and there's no spiny wheel or any messaging, there's just empty space
3. The summary wheel, file formats wheel, and time period graph have a never-ending spiny wheel 

**Other Considerations**
- Is the following issue related to this? Or is this issue completely resolved? I think it might be unrelated/resolved. [metacatui#1927](https://www.google.com/url?q=https://github.com/NCEAS/metacatui/issues/1927&sa=D&source=calendar&ust=1659305690349852&usg=AOvVaw2BUCFe1Z-g-usljBGnDGh_): Portal metrics are missing from recent months
- It seems like either (a) any portal with datasets that were created recently will have at least one issue OR (b) any portal that was submitted recently will have at least one issue. 

<br> 

**To Reproduce**
Check out these portals.

**ESS-DIVE**
- Portals with all three bugs: [watershed sfa](https://data.ess-dive.lbl.gov/portals/watershed-function-sfa/Metrics), [ATWE](https://data.ess-dive.lbl.gov/portals/ATWE/Metrics)
- Portals with no issues: [WHONDRS](https://data.ess-dive.lbl.gov/portals/WHONDRS/Metrics), [SPRUCE](https://data.ess-dive.lbl.gov/portals/SPRUCE_Experiment/Metrics), [CDIAC](https://data.ess-dive.lbl.gov/portals/CDIAC/Metrics)

**Arctic Data Center** 
[...]

PR #2088, opened 2023-02-03
Resolves issues with infinite spinners for portal metrics

rushirajnenuji:
Part of #2041 
MetacatUI was sending URL encoded query strings as part of POST request query parameter which was causing issues with SOLR request resulting in `500` error. The commit in this PR fixes those errors for Portal Metrics.


robyngit (approved):
Nice catch! 🥳

PR #2203, opened 2023-10-17
Adds `enablePublishOnAllNonPublic` configuration

### LINKED ISSUE #2202: Configure MetacatUI to show publish button even if DOI is assigned.
**Describe the feature you'd like**
Adding a new configuration `enablePublishWithExistingDOI` so that when it is true, users can see a "Publish" button in the MetadataView if enablePublishDOI is also enabled, even if a DOI is assigned to the dataset

**Is your feature request related to a problem? Please describe.**
Since some packages could have the use case where they already have a pre-minted DOI, they still would have to go through ESS-DIVE publication process.

helbashandy:
## Description

Adds `enablePublishOnAllNonPublic` configuration. If true, users can see a "Publish" button in the MetadataView on all non-public datasets regardless if a DOI is assigned to the dataset. If false, the default behavior would take place based on the `AppConfig#enablePublishDOI` config.

closes #2202

### DIFF
--- a/src/js/models/AppModel.js
+++ b/src/js/models/AppModel.js
@@ -1080,6 +1080,20 @@ define(['jquery', 'underscore', 'backbone'],
       isJSONLDEnabled: true,
 
       /**
+      * If true, users can see a "Publish" button in the MetadataView on all non-public datasets regardless if
+      * a DOI is assigned to the dataset
+      * If false, the default behavior would take place based on the {@link AppConfig#enablePublishDOI} config.
+      *
+      * This config helps in the case that a member-node publication workflow requires review of the datasets before
+      * they can be public. It also allows datasets to be reviewed for publication regardless of the DOI status,
+      * as some datasets could have a pre-assigned DOI when transferring their dataset to a member node.
+      * @type {boolean}
+      * @default false
+      * @since 2.27.0
+      */
+      enablePublishOnAllNonPublic: false,
+
+      /**
       * If true, users can see a "Publish" button in the MetadataView, which makes the metadata
       * document public and gives it a DOI identifier.
       * If false, the button will be hidden completely.
--- a/src/js/views/MetadataView.js
+++ b/src/js/views/MetadataView.js
@@ -1231,10 +1231,20 @@ define(['jquery',
           }
 
           try {
-            //Determine if this metadata can be published.
+            // Determine if this metadata can be published.
             // The Publish feature has to be enabled in the app.
             // The model cannot already have a DOI
-            var canBePublished = MetacatUI.appModel.get("enablePublishDOI") && !view.model.isDOI();
+            var canBePublished;
+
+            // Check if this metadata can be published based on the "enablePublishOnAllNonPublic" Configuration.
+            // This helps in the case that a member-node publication workflow requires review of the datasets before
+            // they can be public. It also allows datasets to be reviewed for publication regardless of the DOI status,
+            // as some datasets could have a pre-assigned DOI when transferring their dataset to a member node.
+            if (MetacatUI.appModel.get("enablePublishOnAllNonPublic") && view.model.get("isPublic") === false){
+              canBePublished = true;
+            } else {
+              canBePublished = MetacatUI.appModel.get("enablePublishDOI") && !view.model.isDOI();
+            }
 
             //If publishing is enabled, check if only certain users and groups can publish metadata
             if (canBePublished) {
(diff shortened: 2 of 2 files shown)


src/js/models/AppModel.js
```diff
@@ -1079,6 +1079,15 @@ define(['jquery', 'underscore', 'backbone'],
       */
       isJSONLDEnabled: true,
 
+      /**
+      * If true, users can see a "Publish" button in the MetadataView on all non-public datasets regardless if
+      * a DOI is assigned to the dataset
+      * If false, the default behavior would take place based on the {@link AppConfig#enablePublishDOI} config.
+      * @type {boolean}
+      * @default false
```
robyngit, line 1087: When this gets released, I update all the @since x.x.x with the new version.
```suggestion
      * @default false
      * @since x.x.x
```

robyngit (changes requested):
@helbashandy this looks great, thanks! Could you please just update the comments above your code changes to describe the new decision process?

robyngit:
@rushirajnenuji I imagine that your work on hierarchical folders could involve lots of changes to the MetadataView. Does this PR conflict with any of your work?

rushirajnenuji:
Thank you @robyngit  and @helbashandy  -- I'll review the code to see any conflicts with MDV and get back.

rushirajnenuji (approved):
Hi @robyngit  and @helbashandy  -- I reviewed this PR and it looks good. There are some minor conflicts with the MetaDataView, but it should be easy to merge into my branch. Thank you for checking.

src/js/models/AppModel.js
robyngit: That's perfect @helbashandy, thank you! In the future, just use `x.x.x` as the version - these are all replaced with the actual version number at the time of release - that way there's flexibility if something has to get pushed to a different release, or the version # changes for any reason.

robyngit:
@helbashandy not sure what the expected behaviour is for ESS-DIVE, but it seems intuitive that if enablePublishDOI is false, the publish button should never show up. Likewise if publishing is limited to those in the enablePublishDOIForSubjects list, and the user isn’t authorized, then the publish button should never show up. So if we were to move this decision into its own method, the logic would be something like...

```js
canBePublished: function () {

  // The Publish feature has to be enabled in the app. 
  if (!MetacatUI.appModel.get("enablePublishDOI")) {
    return false;
  }

  // If the repository is configured to limit publishing to only certain
  // users and groups, then check if the logged-in user is one of the
  // subjects in the list or is in a group that is in the list. If not,
  // then this metadata cannot be published.
  const authorizedPublishers = MetacatUI.appModel.get("enablePublishDOIForSubjects");
  const authorizedOnly = Array.isArray(authorizedPublishers) &&
    (authorizedPublishers.length);
  const userIsAuthorized = authorizedOnly &&
    MetacatUI.appUserModel.hasIdentityOverlap(authorizedPublishers);
  if (authorizedOnly && !userIsAuthorized) {
    return false;
  }

  // Finally, if the setting to enable publishing on all private
  // metadata is enabled, then allow publishing if the metadata is not
  // yet public. Otherwise, allow publishing if there's not a DOI yet.
  const isPrivate = !this.get("isPublic")
  const isDOI = model.isDOI()
  const enablePublishOnAllPrivate = MetacatUI.appModel.get("enablePublishOnAllPrivate");
  if ((enablePublishOnAllPrivate && isPrivate) || !isDOI) {
    return true;
  }

  // Otherwise, this metadata cannot be published.
  return false;

},
```

What do you think?

robyngit:
Hey @helbashandy, wondering if ESS-Dive is still hoping to merge in these changes and add a enablePublishOnAllNonPublic config option?

robyngit:
I will close this PR for now since it hasn't had attention for 7+ months, but @helbashandy please re-open if you're still wanting to get this merged into MetacatUI!

helbashandy:
Hi Robyn,

I apologize for the delayed response. Regarding the pending PR, we were in the process of finalizing design changes that impact the ESS-DIVE user journey for publication requests. After consideration of our current use-cases, we determined that the suggested configuration change is no longer necessary. 

The current design includes an additional modal triggered by the publication button, which addresses various use cases beyond what the suggested configuration would have covered. As a result, integrating the config change is not currently a priority for ESS-DIVE's UI and user journey flow.

Thank you for your patience and for following up on the PR. Let me know if you have any further questions.

robyngit:
Thanks @helbashandy ! I figured you had taken a different path, but just didn't want to assume :)

PR #2204, opened 2023-10-17
Draw on Cesium map feature

robyngit:
This PR introduces the feature that allows a user to draw an arbitrary polygon on a map. While we don't yet have anything set up to use this custom polygon, the "save" button has the ability to send the coordinates a callback function. This will support us in developing upcoming features: 1) allow users to download a subset of geospatial data, and 2) allow users to plot a subset of the data.

In making these changes, I also did a little refactoring of some of the related models and collections to make them more modular & reusable.


robyngit:
Findings from @artntek 

- Encountered a scenario where the map froze while loading results.
- This freeze happened again when zooming out.
- The issue hadn't been observed previously on the ADC Cesium map (though not much time spent exploring the ADC version)
- The map tab did not crash, but the loading took longer than a minute.
- On moving the browser tab, the expected results eventually displayed.

Screenshot 2023-10-18 at 1 52 27 PM
Screenshot 2023-10-18 at 1 47 56 PM
Screenshot 2023-10-18 at 1 55 47 PM

robyngit:
@rushirajnenuji, I am working on reproducing the problem described above. Other than that, I don't expect to make changes to this PR, so please feel free to start reviewing anytime!

robyngit:
@rushirajnenuji reported a behaviour he encountered where datasets and the catalog keep refreshing on the map ([see video](https://www.loom.com/share/7b72273269ca449d975fb5fbbf274b04?sid=a0c8c49f-0249-41c0-acd2-0628d5066cf2)). He noted that no external triggers seem to cause it, and that he hadn't seen this happen before. This should be addressed before we release this feature.

robyngit:
The issues observed by @rushirajnenuji and @artntek seemed to stem from the Cesium camera events triggering unpredictably (reported [here](https://github.com/CesiumGS/cesium/issues/4753) and [here](https://community.cesium.com/t/multiple-firing-of-movestart-moveend-events-when-getting-camera-heading-pitch-roll/9045)). I added a work around so that instead of relying solely on Cesium's "moveStart" event, we trigger our own "moveStartAndCameraChanged" event that occurs after the camera starts moving and has moved enough to be considered a "changed" camera position. This seems to resolve the issue, although it's very difficult to reproduce so more testing will be needed. This branch at it's current state, which now also includes the develop changes, is deployed at http://metacatui.test.dataone.org/

robyngit:
Going to merge this into `develop` so that I can add a couple of features that depend on the changes made here. Review can be completed along with the main release PR.

PR #2209, opened 2023-10-25
Feature 2079 file display

rushirajnenuji:
status: draft PR


robyngit:
Started reviewing this @rushirajnenuji, looking good so far!! I realize this is just a draft PR so perhaps these items are on your ToDo list already, but a few quick tasks to add if not:
- [ ] JSdocs for all new and modified methods and properties (including `@since x.x.x` tags for the new ones).
- [ ] Tests for all of the new methods in models & collections.

robyngit:
Example data packages with `atLocation` information for testing:
- https://arcticdata.io/catalog/view/doi%3A10.18739%2FA2NK36626
- https://arcticdata.io/catalog/view/doi:10.18739/A2V40K14Q

robyngit:
@rushirajnenuji, I've been testing out the new DataPackageView, and I must say, it looks great! The UI is going to be a big step forward. I did notice some performance issues that we might need to address. Loading data packages with a large number of files seems to take longer using the new view. Using [`doi:10.18739/A2SN4C`](https://arcticdata.io/catalog/view/doi:10.18739/A2SN4C) with 340 files as an example:

- The datatable in metacatui 2.27.0 takes about 9 seconds to load, whereas in `feature-2079-file-display`, it takes around 20-35 seconds. (Sometimes while loading, the browser slows down or the tab nearly crashes.) Any idea what might be causing this extra load time? I imagine something to do with parsing the presumably very large EML doc?
- I also noticed that collapsing the list of files in the datatable takes about 6-10 seconds (expanding it is quick though).

See screencasts below.

For the MVP release, we could temporarily revert to the old data table view for handling very large packages. It's a bit of a workaround, considering our goal is to enhance support for larger datasets, but it might help maintain performance while we fine-tune the new feature. What do you think about this? Should we consider delaying the release to sort this out, or do you have other suggestions? Really appreciate all the effort you've put into this!!

https://github.com/NCEAS/metacatui/assets/26600641/ed0a9480-7bde-4101-8e51-4c0566b97fb7

https://github.com/NCEAS/metacatui/assets/26600641/ab97e376-0b41-4b80-82d0-f63495f675a6

rushirajnenuji:
Thanks for testing the new DataPackageView and sharing your insights, @robyngit! Your feedback has been crucial for improvement.

current status: I'm fixing some algorithms introduced for these features and aim to have an update ready by tomorrow's dev call. We can also discuss the release plan on the call.

robyngit:
@rushirajnenuji just added one commit with some very minor CSS/layout tweaks that:
- keeps the table header border visible when you scroll down
- removes the "Files in this dataset" row from the edit version of the view
- adds a `<tr>` around the header elements in edit mode so it inherits the same CSS rules as the non-edit mode - specifically, keeps the header background white so you can still read the header text when you scroll down

Before:
Screenshot 2024-02-01 at 15 07 18

After:
Screenshot 2024-02-01 at 15 45 48

robyngit:
So far it's all looking great @rushirajnenuji, except that I can't get the package table to work in the dataone theme, it stays stuck on the `Retrieving data set details...` mode:

Screenshot 2024-02-01 at 15 47 29

I let it load for about 10 minutes and it never moved past that screen. If I switch back to the main branch then it loads as expected, so I believe it's something to do with the new changes. To reproduce the issue, just switch to the dataone theme and open any dataset in view mode. Any idea what might be causing this?

rushirajnenuji (commented):
Hey @robyngit  -- thank you for catching and fixing this. I tested this changes on my local instance and they look good. (sorry looks like github doesn't let me approve this, so I'm leaving a comment here)

rushirajnenuji:
re: _dataone theme_
I believed it worked in dataone theme as well, not sure if the most recent changes broke something. I'll take a look and try to add a fix for this ASAP. Thank you for catching this Robyn, I really appreciate all your time and help with getting this feature **production** ready.

robyngit (approved):
Thanks to @justinkadi and Maggie for help reviewing!

Issue #2243, opened 2024-01-16
Improve the NSF awards API endpoint


robyngit:
We got the awards API working again in issue #2233, however, there are some peculiarities with how the API works that we should address to make it more useful.

As explained in the [NSF awards API documentation](https://resources.research.gov/common/webapi/awardapisearch-v1.htm), the API can be searched with a keyword that is used in a "free text search across **all the available awards data**". The awards data includes many fields that are not useful for our purposes, and there is currently no way to search only within the title.

To work around this, the LookupModel filters the results from the API to include only those that have the keyword in the title. Since the API only returns a maximum of 25 results, this means that if the keyword is not in the title of any of the **first** 25 results, the autocomplete will show no results. There may be results after the first 25 that have the keyword in the title, but we don't attempt to find them.

Some possible solutions:
1. Request that the NSF awards API be updated to allow searching only within the title.
2. Send multiple requests to the API. After each request, filter the results to include only those that have the keyword in the title. If there are fewer than 25 results, send another request for the next 25 results. Repeat until there are 25 results or there are no more results. This downside of this approach is that it will take longer to get the results, and it will require more requests to the API.
3. Don't filter the results at all. This would allow the autocomplete to show results that don't have the keyword in the title, but from experimentation I feel that this will return less relevant results.

While working on this, we should consider externalizing all of the funding API logic to it's own model. It's currently in the LookupModel, but it's not making use of any other LookupModel functionality, and the LookupModel is already quite large. Solving this issue could also add complexity which would further justify moving encapsulating the funding API logic in it's own model.

PR #2244, opened 2024-01-17
Fix the NSF award API endpoint

robyngit:
The major part of the fix involved setting up the ADC proxy to point to the new NSF grants API endpoint, then using the proxy in MetacatUI rather than the endpoint directly since it doesn't support CORS. Other minor fixes and improvements made along the way include:

- Remove unused getGrant method from the LookupModel
- Use native fetch API instead of jQuery.ajax to retrieve grants
- Show spinner to indicate that grants are being retrieved (was set up but not used)
- Move some logic from the getGrantAutocomplete method to new, smaller methods
- Add missing JS docs
- Add error handling
- Add a guide to help repos set up their own proxy
- Start tests for the LookupModel, including a test for the getGrants method


robyngit:
@rushirajnenuji, no rush, but if would you have a chance to give this a quick test to make sure everything's ok? You should only need to set the theme to "arctic" in order for the funding autocomplete field to work in the editor.

rushirajnenuji:
Thank you @robyngit  -- I'll test it out and get back to you.

robyngit:
@rushirajnenuji forgot to point out that I did identify some improvements that are needed to this in the future: https://github.com/NCEAS/metacatui/issues/2243 (just to save you some time in case you come to the same conclusions)

rushirajnenuji (approved):
Reviewed these changes on my local instance of MetacatUI and they look good. Thank you for getting this fix in @robyngit!

robyngit:
Thanks @rushirajnenuji !

Issue #2242, opened 2024-01-08
Add scroll bar to the cesium layer list


robyngit:
In a Cesium map, when there are too many layers in in the list, the layer list expands beyond the window height, expanding the height of the page/map with it. Until we support nested data layers (#1773) or other UI elements to organize long lists of layers, the layer list should at least have a scrollbar when it's taller than the window height.

PR #2249, opened 2024-01-20
Add CSS to support scrolling in layers toolbar in the Cesium map view

ianguerin:
- Set a fixed height for body element to force map view to appear on page.
- Add some padding to allow for scrollbar in layers toolbar.
- Fix issue with cutting off last layer in toolbar on narrow laptop screen due to a hardcoded value for the header height.

GitHub issue: https://github.com/NCEAS/metacatui/issues/2242

### DIFF
--- a/src/css/map-view.css
+++ b/src/css/map-view.css
@@ -180,7 +180,7 @@
   --headerHeight: 80px;
   position: relative;
   height: 100%;
-  min-height: calc(100vh - var(--headerHeight));
+  min-height: 100%;
   width: 100%;
   display: grid;
   box-sizing: border-box;
@@ -229,6 +229,8 @@
   top: 0;
   left: 0;
   height: 100%;
+  overflow: hidden;
+  width: min-content;
   /* required to be placed above map widget in firefox: */
   z-index: 1;
 }
@@ -395,6 +397,7 @@ represents 1 unit of the given distance measurement. */
   class .toolbar, is not hidden so that the links (tabs) are still visible when the
   toolbar is closed */
   display: none;
+  overflow: hidden;
 }
 
 .toolbar--open .toolbar__all-content {
@@ -408,10 +411,12 @@ represents 1 unit of the given distance measurement. */
   justify-content: center;
   /* hide unless the content section is active */
   display: none;
+  overflow: auto;
 }
 
 .toolbar__content--active {
   padding-top: 1rem;
+  padding-right: 1rem;
   display: flex;
 }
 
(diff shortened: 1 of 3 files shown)


mbjones:
Thanks @ianguerin this is great. Could you provide a screenshot in the PR so folks that don't have a dev-instance of MetacatUI set up can see what the feature will look like?

@robyngit could you review please and maybe this can be included in the release along with the hierarchical data view? Thanks.

ianguerin:
Here are some before an after screenshots:
Before the layers list was scrollable:
before being scrollable

After, with the layers list scrollable
after_scrollable

Before, when scrolled to the bottom:
before, scrolled to the bottom

After, when scrolled to the bottom
after, scrolled to the bottom

robyngit (changes requested):
Great work, @ianguerin! Thank you for being so proactive on this and congrats on your first MetacatUI PR :tada:! Your changes look perfect for the PDG portal in the arctic theme. I appreciate that you added only the most minimal CSS rules to make this work; it helps keep our already very large CSS file from getting even larger!

Because MetacatUI is used by other projects, it's important for us to check that changes align well across various themes and contexts before merging them into the main codebase. I've outlined some of these considerations below.

### In non-arctic themes

When testing in non-arctic themes, like knb, there seems to be a slight issue: the cesium map expands to accommodate this list, instead of having the list scroll as intended.

To switch to another theme, change `theme: "arctic"` to `theme: "knb"` in the `config.js` file.

### In non-portal contexts

Repositories may also use the cesium map, instead of Google maps, in the main data search catalog. In this context, there's a bit of a display problem. The `overflow: hidden` CSS rule added to the `.map-view__toolbar-container` class is causing the right side of the layer list to be clipped off.

To see this, pull in the `drp` theme from the [metacatui-themes](https://github.com/NCEAS/metacatui-themes/) repo, and configure the `theme` option to `drp`. The search catalog with layer list expanded looks like this:

Screenshot 2024-01-22 at 16 20 24


---

We can collaborate on fixing this issue given that other themes and contexts may be outside the scope of your PDG work. In any case, thank you again for this and I look forward to more of your contributions!

ianguerin:
Thanks for the detailed review, Robyn. I've updated the PR with changes to address the issues you found with the toolbar. It looks like the issue with the `drp` theme was probably broken without any scroll capability (you couldn't actually access the layers in the list below the visible content), so this change should help.

robyngit (approved):
Nice work @ianguerin!!

Issue #2207, opened 2023-10-23
Toolbar menu blocks clicking on Cesium Map


robyngit:
As reported by @rushirajnenuji, the div that holds the toolbar menu in Cesium is placed such that none of the map behind it is clickable. The div runs the entire height of the map but contains only a handful of buttons, resulting in a large section of the visible map being unclickable. This is very apparent when drawing polygons (#2180). This issue can be fixed with a little CSS finessing.

![Screen Shot 2023-10-23 at 3 12 01 PM](https://github.com/NCEAS/metacatui/assets/26600641/3b25fe43-0a51-458f-9b77-ce079ddfcce0)

PR #2255, opened 2024-01-29
Fit cesium map toolbar size to content

ianguerin:
At 100% height the toolbar blocked the map below even though it had no interactable elements below.

GitHub issue: #2207

fix-bug-toolbar-blocks-content

### DIFF
--- a/src/css/map-view.css
+++ b/src/css/map-view.css
@@ -228,7 +228,7 @@
   position: absolute;
   top: 0;
   left: 0;
-  height: 100%;
+  height: max-content;
   /* required to be placed above map widget in firefox: */
   z-index: 1;
 }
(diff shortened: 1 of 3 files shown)


robyngit:
Hey @ianguerin, thanks for picking up another issue! Wondering why you closed this PR? I think this is the preferable solution over https://github.com/NCEAS/metacatui/pull/2256

ianguerin:
I didn't intend to open this PR, that was my mistake while wondering "what does this button do" on GitHub. 
The issue with this approach is that the bug persists when you open the layers list up again, since the size of the container is again 100% of the height of the page.
blocks-when-open

robyngit:
ah, understood, thanks!

Issue #2246, opened 2024-01-17
Fly to coordinates in Cesium map


iannesbitt:
**Describe the feature you'd like**
DRP team members have requested a way to fly to specific coordinates in the Cesium map. This could exist as a button that reveals a `lat, lon` text entry field (or two separate fields), which is then used to change the map focus location in a fly to manner similar to that of the behavior after pressing the Home button.

**Is your feature request related to a problem? Please describe.**
Researchers may be interested in a specific area of the map, which may be easiest to find with a copy/paste of coordinates. This would help researchers and other users zoom to areas of interest quickly.

**Additional context**
n/a

robyngit:
I think this would be one of the more straight forward features to implement. Here are some pointers:

- The Map Model already has [a "zoomTo" method](https://github.com/NCEAS/metacatui/blob/758ff8e3710637d9fdf406a9c2383e56ff5be877/src/js/models/maps/Map.js#L269-L280) that this widget could call when the coordinates are entered into the entry field
- We could potentially make use of the [GeoPoint model](https://github.com/NCEAS/metacatui/blob/main/src/js/models/maps/GeoPoint.js) to validate the inputted coordinates.
- The widget should be its own independent view that interacts with the Map Model (e.g. see the [ToolbarView](https://github.com/NCEAS/metacatui/blob/main/src/js/views/maps/ToolbarView.js))
- The [MapView](https://github.com/NCEAS/metacatui/blob/main/src/js/views/maps/MapView.js) should do the work of creating an instance of this widget and inserting it into the map interface.
- We should design this feature with https://github.com/NCEAS/metacatui/issues/1796 in mind. For example, the widget we build here could eventually be extended to navigate to not only specific coordinates, but also place names or addresses.

iannesbitt:
Thanks Robyn! I agree that it would be good to include place names as part of the search functionality. I will take a look at this when I have the opportunity.

ianguerin:
Here's a demo of what I have built so far: https://drive.google.com/file/d/1Sukzcv3BjcxH3Pi2fNHLd1g2Oii7LNu7/view?usp=drive_link

I'm working on some unit tests now, but figured I'd share this in case it doesn't match expectations

robyngit:
Looks great @ianguerin! @iannesbitt any comments from the DRP perspective?

PR #2258, opened 2024-02-02
Feature #2246- view finder feature to fly to user-specified latitude, longitude pair

ianguerin:
This feature allows the user to input two floating point numbers, with helpful error messages when they enter malformed queries, or values that are out of bounds (-90:90, -180:180), or non-numerics.

The input becomes focused when the "view finder" layer section is opened.

This PR also introduces two extremely slimmed down "libraries" that use implement some best practices for unit tests (hermetic state, function spies) without adding additional third party dependencies.

fixes NCEAS/metacatui#2246


ianguerin:
Here's a demo video of the feature: https://drive.google.com/file/d/1Sukzcv3BjcxH3Pi2fNHLd1g2Oii7LNu7/view?usp=drive_link

@robyngit, there's a lot of code here and I expect some discussion about it! Please let me know where I strayed too far from the current codebase style

robyngit:
From a quick scan of your changes @ianguerin, it looks fantastic! I'll share more feedback after I've had a chance to go through this more thoroughly.

@iannesbitt, if you have the time, could you please test out these changes in the DRP theme and make sure it meets that group's requirements?

src/js/views/maps/ViewFinderView.js
```diff
@@ -0,0 +1,155 @@
...
+     * @typedef {Object} ViewFinderViewOptions
+     * @property {Map} The Map model associated with this view allowing control
+     * of panning to different locations on the map. 
+     */
+    initialize(options) {
+      this.mapModel = options.model;
+    },
+
+    render() {
+      this.focusInput();
```
yvonnesjy, line 62: Not familiar with the backbone life cycles yet- are we calling focusInput both here and in action() because render is only called once, but action is called every time user navigates to that tab?

src/js/views/maps/ViewFinderView.js
```diff
@@ -0,0 +1,155 @@
...
+
+      this.el.innerHTML = _.template(Template)(this.templateVars);
+    },
+
+    /**
+     * Focus the input field on a delay to allow for the input to be
+     * visible on the page before attempting to focus.
+     */
+    focusInput() {
+      _.defer(() => {
```
yvonnesjy, line 72: Just curious- what's the error case you ran into here that prompted _.defer?

src/js/views/maps/ViewFinderView.js
```diff
@@ -0,0 +1,155 @@
...
+        const input = this.getInput();
+        input.focus();
+        // Move cursor to end of input.
+        input.val("");
+        input.val(this.templateVars.inputValue);
```
yvonnesjy, line 77: I'm trying to figure out why you created this.templateVars.inputValue, instead of just using this.$el.find(`.${classNames.input}`). It seems like this block might be why... but I still don't fully understand. Can you explain?

src/js/views/maps/ViewFinderView.js
ianguerin: `ToolbarView` actually renders each of it's sections' corresponding views immediately, and then shows and hides them. Unfortunately that means that when the input becomes visible, the `render` function has already been called. The `render` function is called manually and it seems that most views call it in their `initialize` function which is similar to a `constructor` function.

src/js/views/maps/ViewFinderView.js
ianguerin: The main reason that I need to track the input value is so that the UI remains consistent across re-renders. So in the template I also reference this value, otherwise the input would be cleared each time I manually re-render. 
I think your questions makes it clear that there should be a comment to that effect where the `templateVars` object is initialized

src/js/views/maps/ViewFinderView.js
yvonnesjy: nit: should we remove the focusInput() here if it's gonna be called again in action() anyway?

src/js/views/maps/ViewFinderView.js
```diff
@@ -0,0 +1,155 @@
...
+      return { latitude, longitude };
+    },
+
+    clearError() {
+      this.setError("");
+    },
+
+    setError(errorMessage) {
+      this.templateVars.errorMessage = errorMessage;
+      this.render();
```
yvonnesjy, line 137: are we calling render() here because setting templateVars doesn't update the html automatically?

src/js/views/maps/ViewFinderView.js
yvonnesjy: Does calling focusInput() after `this.el.innerHTML =` work?

iannesbitt (changes requested):
Hi @ianguerin—thank you for diving in and creating a working feature so quickly! The Cesium visualization is in desperate need of this feature, and the clients we work with will be very pleased to be able to use it.

This feature is overall very smoothly integrated. I have not been able to break it or cause anything ugly to happen, which is awesome. I very much appreciate the heads up to the user when they enter coordinates outside of the `[-90,90], [-180,180]` ranges. I do however have a couple of items to address—one that's pressing and two that would be nice-to-have.

### Zoom level

It seems to me that the zoom level is set too high when entering a point. My guess is that when a user enters a point, they'd like to see a much lower (tighter) zoom on that point on the Cesium globe. The research team that requested this feature, for example, will want to zoom directly to their study sites. This zoom level feels like it will still require them to do some work to find their site.

One possibility would be to zoom tighter based on the number of decimals supplied by the user, so for example entering `42,-71` might give you a zoom level that includes most of the state of Massachusetts, whereas `42.7, -73.1` might zoom to the town of North Adams. A calculation that makes this possible in a 2-D map wouldn't be too hard I don't think—we could just zoom to a bounding box that includes ± the last decimal values in the query to get rough extents to zoom to. This might be somewhat harder in an application like Cesium, however, and may need its own feature request depending on whether others think this is a good idea.

To summarize: zoom level should be much lower at least, and at most may be able to be calculated based on the decimal precision.

### Place-based search

This is more of a "reach" feature that we could consider adding in the future. However, there are several geocoder (place name/location) databases out there that we could piggyback to be able to allow place name searches (see [OSMNames](https://osmnames.org/about/)). For example, a search for `New York, NY` would return a coordinate pair or bounding box and then we could have Cesium center the map on roughly that location (`40.7, -74.0`).

To summarize: we should include a way to search by place names (in the future).

### Conclusion

I will approve the PR when the zoom level is tightened, but we should also consider the following additions to accompany this feature:

- [ ] Zoom level based on decimal precision
- [ ] Search based on place name database

Thank you once again for submitting this PR, this is very useful already and represents a big step forward for MetacatUI-Cesium UX possibilities. I'm incredibly excited to use it myself and to introduce our clients to it!

src/js/views/maps/ViewFinderView.js
ianguerin: Action is *only* called when the user clicks the magnifying glass button to make this "View finder" panel visible; so only called once. After the user types "enter" or clicks the search button (on the input) I want the input to stay focused, so we need to call focus after each render.

src/js/views/maps/ViewFinderView.js
ianguerin: Yep! Good call

src/js/views/maps/ViewFinderView.js
ianguerin: Correct, there's no concept of bound variables as far as I can tell. Typically you'd register the render function as an event listener on a Model's change event. However, this templateVars object is not a Model, and I don't necessarily want to re-render every time it changes anyhow (like when I'm tracking the input's value)

ianguerin:
Thanks for your review @iannesbitt, I've updated the zoom to 10km instead of 321km, hopefully that is a more useful starting point. 
While I do think searching for places is a lot more useful, I think that belongs as a separate feature and development process (one that I'm actively researching now) because of the work it will take to select a geocoding service, and also build out the UI for displaying a potential multitude of results corresponding to a place name.

@robyngit, it came up today whether or not it would be a good idea to release this feature _without_ support for place name search.  I'm curious to hear your thoughts, but also curious if there is a way we can put this feature behind a feature flag for now, or if we even have support for enabling/disabling features flags.

robyngit:
@yvonnesjy @iannesbitt thank you for adding your reviews, very helpful!! 🎉 

@ianguerin thank you for being thorough and adding tests! I think they were passing initially but it looks like something changed? `npm test` gives 4 failures now:

```
[1]: ViewFinderView Test Suite
     > bad search queries
     > AssertionError: expected '' to contain 'Try entering a search query with two …'
    at Context.<anonymous> (js/specs/unit/views/maps/ViewFinderView.spec.js?v=undefined:110:50)
[2]: ViewFinderView Test Suite
     > bad search queries
     > AssertionError: expected '' to contain 'Try entering a search query with two …'
    at Context.<anonymous> (js/specs/unit/views/maps/ViewFinderView.spec.js?v=undefined:128:50)
[3]: ViewFinderView Test Suite
     > bad search queries
     > AssertionError: expected '' to contain 'Latitude values outside of the'
    at Context.<anonymous> (js/specs/unit/views/maps/ViewFinderView.spec.js?v=undefined:146:50)
[4]: ViewFinderView Test Suite
     > bad search queries
     > AssertionError: expected '' to contain 'Longitude values outside of the'
    at Context.<anonymous> (js/specs/unit/views/maps/ViewFinderView.spec.js?v=undefined:164:50)
::error::One or more MetacatUI tests failed. Test failure details can be viewed by running "npm view-tests". %0AFAILS:  4%0APASSES: 315%0AFailed Tests: %0A-------------%0A%0A[1]: ViewFinderView Test Suite%0A     > bad search queries%0A     > AssertionError: expected '' to contain 'Try entering a search query with two …'%0A    at Context.<anonymous> (js/specs/unit/views/maps/ViewFinderView.spec.js?v=undefined:110:50)%0A[2]: ViewFinderView Test Suite%0A     > bad search queries%0A     > AssertionError: expected '' to contain 'Try entering a search query with two …'%0A    at Context.<anonymous> (js/specs/unit/views/maps/ViewFinderView.spec.js?v=undefined:128:50)%0A[3]: ViewFinderView Test Suite%0A     > bad search queries%0A     > AssertionError: expected '' to contain 'Latitude values outside of the'%0A    at Context.<anonymous> (js/specs/unit/views/maps/ViewFinderView.spec.js?v=undefined:146:50)%0A[4]: ViewFinderView Test Suite%0A     > bad search queries%0A     > AssertionError: expected '' to contain 'Longitude values outside of the'%0A    at Context.<anonymous> (js/specs/unit/views/maps/ViewFinderView.spec.js?v=undefined:164:50)
```

Do you know what changed?

ianguerin:
Thanks for taking a look @robyngit, I have the tests passing locally and have just pushed those changes. 
I've also renamed things from "view finder" to "viewfinder", as well as adding a a map config value to decide whether or not to show this new UI (hidden by default). I've refactored the code in ToolbarView to make that logic a bit more reusable as I noticed all of the buttons in that toolbar had the same kind of logic.

I'm going to look into adding a pre-commit hook that checks to see if unit tests are all passing before committing changes. This should also be helpful in the future if we decide to add linting!

src/js/templates/maps/viewfinder.html
```diff
@@ -0,0 +1,19 @@
+<div class="<%= classNames.baseClass %>">
+  <h2>Viewfinder</h2>
+  <span>Search for a latitude and longitude pair and click the "search" button to show that point on the map.</span>
```
robyngit, line 3: Minor detail:

```suggestion
  <span>Enter a latitude and longitude pair and click search to show that point on the map.</span>
```

test/js/specs/shared/clean-state.js
```diff
@@ -0,0 +1,39 @@
...
+   * Example usage:
+   * 
+   * const state = cleanState(() => {
+   *   const someClassInstance = new ClassToBeUsedInTest();
+   *
+   *   return { someClassInstance };
+   * }, beforeEach);
+   * 
+   * Now state.someClassInstance can be used with a guarantee that it won't leak
+   * state from test to test.
```
robyngit, line 20: 🤯 love it!!!!

test/js/specs/unit/views/maps/ViewfinderView.spec.js
```diff
@@ -0,0 +1,199 @@
+define([
+  "views/maps/ViewfinderView",
+  "models/maps/Map",
+  // The file extension is required for files loaded from the /test directory.
```
robyngit, line 4: 🎉 omg thank you

test/js/specs/unit/views/maps/ViewfinderView.spec.js
```diff
@@ -0,0 +1,199 @@
...
+    }, beforeEach);
+
+    it("creates a ViewfinderView instance", () => {
+      state.view.should.be.instanceof(ViewfinderView);
+    });
+
+    it("has an input for the user's search query", () => {
+      state.view.render();
+
+      state.harness.typeQuery("123")
```
robyngit, line 29: This is so cool and will be super useful for writing tests for other views!!

src/js/views/maps/ViewfinderView.js
```diff
@@ -0,0 +1,154 @@
...
+      const hasBannedChars = value.match(bannedCharactersRegex) != null;
+      if (matches?.length !== 2 || isNaN(matches[0]) || isNaN(matches[1]) || hasBannedChars) {
+        this.setError("Try entering a search query with two numerical values representing a latitude and longitude (e.g. 64.84, -147.72).");
+        return;
+      }
+
+      const latitude = Number(matches[0]);
+      const longitude = Number(matches[1]);
+      if (latitude > 90 || latitude < -90) {
+        this.setError("Latitude values outside of the range of -90 to 90 may behave unexpectedly.");
```
robyngit, line 122: Curious, how does it behave? Is the behaviour ever useful?

src/js/views/maps/ViewfinderView.js
```diff
@@ -0,0 +1,154 @@
...
+    },
+
+    /** 
+     * Values meant to be used by the rendered HTML template.
+     */
+    templateVars: {
+      errorMessage: "",
+      // Track the input value across re-renders.
+      inputValue: "",
+      placeholder: "Search by latitude and longitude",
```
robyngit, line 50: Consider showing an example lat and long instead

src/js/views/maps/ViewfinderView.js
```diff
@@ -0,0 +1,154 @@
...
+
+      this.model.zoomTo({ ...coords, height: 10000 /* meters */ });
+    },
+
+    /**
+     * Parse the user's input as a pair of floating point numbers. Log errors to the UI
+     * @return {{Number,Number}|undefined} Undefined represents an irrecoverable user input,
+     *   otherwise returns a latitude, longitude pair.
+     */
+    parseValue(value) {
```
robyngit, line 111: _Generally speaking_, parsing and validation of values belongs in backbone models rather than views, and we do have a GeoPoint model that we could add parsing to and use here. However, this is a pretty simple case, and so no reason to slow down getting this feature into develop! More of an FYI than a suggestion for this PR.

src/js/views/maps/ViewfinderView.js
```diff
@@ -0,0 +1,154 @@
...
+    initialize(options) {
+      this.mapModel = options.model;
+    },
+
+    render() {
+      this.el.innerHTML = _.template(Template)(this.templateVars);
+
+      this.focusInput();
+    },
+
```
robyngit, line 68: I noticed that some methods here are missing JSDoc comments. I realize these methods are quite straightforward, but just some very brief JSDoc descriptions would be great. It not only helps less experienced developers in the future, but also helps keep our [API docs](https://nceas.github.io/metacatui/docs/) complete.

robyngit (changes requested):
I was so happy to see how you managed to write these tests for views!! I'll be making use of this strategy moving forward. The comments and suggestions I added are all super minor and I did not manage to break the view finder no matter what I entered in the search box. I appreciate that you worked it into the Toolbar following the existing pattern. Very nice work!

The only other comment to add is about docs in Toolbar View on [this line](https://github.com/NCEAS/metacatui/pull/2258/files#diff-de0d1dd6efbe17ef7f9758077d7a68916d4e7c37ad34c441aee1102a3c9968e0R469-R471). `* @returns {HTMLElement} Returns the content container with the rendered view` just needs to be updated to account for the change in what's returned from the method:

```
         * @returns {Object} Returns an object with the content container and the view
         * that was rendered in that container.
         * @property {HTMLElement} contentContainer The container for the toolbar section's
         * content
         * @property {Backbone.View} sectionContent The view that was rendered in the
         * content container
```

(GitHub wouldn't allow me to comment directly since this isn't one you made changes to).

src/js/views/maps/ViewfinderView.js
ianguerin: Done. Added JSDoc style comments to each method and field.

src/js/views/maps/ViewfinderView.js
ianguerin: As this feature gets more functionality (like geocoder integration) I think it will be helpful to have the examples of possible search queries live in the "error" messaging.

src/js/views/maps/ViewfinderView.js
ianguerin: It behaves appropriately as far as I can tell. I mean this error message mostly to make sure that the user meant to enter the value that they entered. For example if they accidentally type "47, -1701.2" instead of "47, -170.12", it's likely that they should double check their input even though the Cesium viewer will take them to -1701.2 mod 180

test/js/specs/unit/views/maps/ViewfinderView.spec.js
ianguerin: I noticed there are no *View.spec.js files, so hopefully this is a useful pattern to get us started. Using a "harness" is a great way to hide the implementation details, especially when writing tests for reusable components. this way if the class name of something changes, we'll only need to update the harness, rather than every single test that uses the reusable view.

src/js/templates/maps/viewfinder.html
ianguerin: Done.

src/js/views/maps/ViewfinderView.js
```diff
@@ -0,0 +1,154 @@
...
+  Template,
+) => {
+  /**
+   * @class ViewfinderView
+   * @classdesc The ViewfinderView allows a user to search for a latitude and longitude in the map view.
+   * @classcategory Views/Maps
+   * @name ViewfinderView
+   * @extends Backbone.View
+   * @screenshot views/maps/ViewfinderView.png
+   * @since 2.27.1
```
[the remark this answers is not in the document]
ianguerin: Great I'll set as x.x.x.
I'm curious, do we actually find that @since tagging useful? It sounds like undue burden on an already fairly manual release process

src/js/views/maps/ViewfinderView.js
ianguerin: I'd like to get this right. Even if I make the fix in a later commit. 
What would you expect this to look like? 
I'm guessing:
This view instantiates a GeoPoint
GeoPoint is capable of parsing a string and setting its own attributes appropriately

Does that work?

src/js/views/maps/ViewfinderView.js
```diff
@@ -0,0 +1,154 @@
...
+
+// Regular expression matching everything except numbers, periods, and commas.
+const bannedCharactersRegex = /[^0-9,.+-\s]/g;
+
+// Class names that correspond to elements in the template.
+const classNames = {
+  baseClass: 'viewfinder',
+  button: "viewfinder__button",
+  input: "viewfinder__input",
+};
```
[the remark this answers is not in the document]
ianguerin: I'd like to say yes, unfortunately fields like `event` and `className` cannot reference this.classNames so I don't think it would work. If I instead use a JS class then this should be possible. Can we use `class` in the repo?

src/js/views/maps/ToolbarView.js
```diff
@@ -183,6 +193,19 @@ define(
...
           //   viewOptions: {}
           // },
+          {
+            label: 'Viewfinder',
+            icon: 'search',
+            view: ViewfinderView,
+            action(view, model) {
+              const sectionEl = this;
+              view.defaultActivationAction(sectionEl);
+              sectionEl.sectionView.focusInput();
```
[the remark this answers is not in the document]
ianguerin: I think it would be more typical (I have little experience with Backbone, so I'm speaking from general UI framework experience) for the view to be rendered when it was activated/shown.  That way we don't run code that we don't need to. If it was expensive to load one of these views, for example, we wouldn't want to load it unless the user requested it. 

I could see there being an option for actionWithDefault or something to that extent. Conversely, we could add logic that says "if there is an action and a view defined, then run the default action then call the custom action". I think that's the most developer friendly solution

src/js/views/maps/ViewfinderView.js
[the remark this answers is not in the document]
robyngit: `events` can also be a method so you can reference `this.classNames` within events like so:

https://github.com/NCEAS/metacatui/blob/main/src/js/views/maps/LayerNavigationView.js#L68-L79

src/js/views/maps/ViewfinderView.js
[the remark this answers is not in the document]
robyngit: Yes, `@since` tags are useful to track when changes were made. It's really a very quick process to switch the `x.x.x`s to the current release version.

src/js/views/maps/ToolbarView.js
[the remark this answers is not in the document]
robyngit: Thanks, that's super useful. Made an issue to address these changes here: https://github.com/NCEAS/metacatui/issues/2269

src/js/views/maps/ViewfinderView.js
robyngit: Interesting! My initial expectation was for the viewfinder to show an error message and _not_ zoom when the coords are outside the valid range. Maybe we could ask the researchers & users if there's ever a case where they would want to enter outside the 90/-90 180/-180 range? Maybe we could indicate somehow the corrected lat,lon that the map actually navigates to in these cases?

src/js/views/maps/ViewfinderView.js
robyngit: yes, that sounds good! See the GeoPoint model [here](https://github.com/NCEAS/metacatui/blob/758ff8e3710637d9fdf406a9c2383e56ff5be877/src/js/models/maps/GeoPoint.js#L4). You could move your parsing logic there, keeping in mind that this model may eventually parse other types of inputs as well (e.g. EML). You could also use the validate method. It might be useful to change the validate method to return not only a string, but also the model attribute that's invalid. A pattern we use in other model validate methods in metacatui is to return an object in the format: 

```js
{
  latitude: "user friendly message about latitude error",
  otherAttrWithError: "another error message",
  ...
}
```

src/js/views/maps/ViewfinderView.js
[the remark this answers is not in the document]
ianguerin: Done

src/js/views/maps/ViewfinderView.js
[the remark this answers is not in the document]
ianguerin: I would say that's what the git commit history is for, but I'm not familiar with who is referencing these tags, so I'll leave that alone!

src/js/views/maps/ViewfinderView.js
ianguerin: I will do this in a follow up commit

ianguerin:
All tests pass locally so I'm not sure what to make of: https://github.com/NCEAS/metacatui/actions/runs/7906483230/job/21581405581?pr=2258

robyngit:
@ianguerin not sure what that fail was all about, but I reran the tests with success :tada:

ianguerin:
> @ianguerin not sure what that fail was all about, but I reran the tests with success 🎉

Super, anything needed before merging?

robyngit:
Ok I made an issue for you to address the out-of-range coordinates behaviour here: https://github.com/NCEAS/metacatui/issues/2270, which can happen in a later release. Otherwise it looks like all concerns were addressed, nice work! :) Looking forward to seeing this in action.

robyngit (approved):

Issue #2265, opened 2024-02-08
`.PortalView` element needs to have height ≠ 100%


iannesbitt:
**Describe the bug**
A [recent change](https://github.com/NCEAS/metacatui/blame/cd2e5f6834e128f0b5a4786e6893e0b1b0f2d866/src/css/metacatui-common.css#L4879) set the `.PortalView` `height: 100%`, I imagine to keep Cesium pages limited to 100% height. This causes problems on portal pages that are longer than 100%, however, as the footer ends up floating over the body text (see screenshot). To solve this, the `height: 100%` definition needs to be moved to a CSS element that is only applied to pages with a Cesium map.

**To Reproduce**
1. Enable `develop` branch in metacatui
2. Enable `arctic` or `drp` theme
3. Go to a portal with a long markdown page (`localhost:3000/portals/permafrost/About` or `localhost:3000/portals/vt`) depending on the theme
4. Scroll down past footer
5. See footer floating above text

**Expected behavior**
The footer should fall below text.

**Screenshots**
![image](https://github.com/NCEAS/metacatui/assets/18689918/7567738a-88b1-4154-9c9b-23f869aa2040)

**Desktop (please complete the following information):**
all

**Smartphone (please complete the following information):**
all

**Additional context**
n/a

iannesbitt:
It looks like this issue occurs because there is no class that differentiates Cesium portal pages from Markdown pages. I think a fix for this that keeps both the Cesium layer list scroll bar functionality and also pushes the footer down the page would have to involve using JS to add or substitute a class to the `<body>` definition that would indicate the page type. Then that class could be assigned a `height: 100%` or `height: unset` based on needs.

iannesbitt:
Although perhaps there is a way to unset the body height on the Markdown pages directly via script tags...?

iannesbitt:
Putting `body {height:100%}` definitions in `map-view.css` or in the Cesium HTML templates ends up causing this problem too. I think this needs a JS solution.

iannesbitt:
I've tried a few things `PortalVisualizationsView`. I can get `height: 100%` added to the style of `<body>` but I can't figure out how to remove it when the user navigates away from the map page to another portal page. The following works in [this `else if` clause](https://github.com/NCEAS/metacatui/blob/develop/src/js/views/portals/PortalVisualizationsView.js#L134-L137) to add the style definition:
```js
              else if( this.model.get("visualizationType") == "cesium" ){
[...]

PR #2271, opened 2024-02-16
Bugfix #2265 portals body height

iannesbitt:
From [my comment](https://github.com/NCEAS/metacatui/issues/2265#issuecomment-1949332448) on the issue:

> My solution to this is a little different than the [above method](https://github.com/NCEAS/metacatui/issues/2265#issuecomment-1936748651), instead of using the `.PortalView` element, I set `height: 100%` on `document.style.body` in `PortalVisualizationsView`, and remove the property in `PortalSectionView`. I did this because it echoes the way [the `--footer-height` is set and unset](https://github.com/NCEAS/metacatui/blob/758ff8e3710637d9fdf406a9c2383e56ff5be877/src/js/views/FooterView.js#L30-L42). If it would be cleaner for me to do this another way (such as setting this on `.PortalView`), I'm happy to switch to a different method.

### DIFF
--- a/src/css/metacatui-common.css
+++ b/src/css/metacatui-common.css
@@ -4879,7 +4879,6 @@ i.toc-sub-item {
 ********************************************/
 
 .PortalView {
-  height: 100%;
   display: grid;
   grid-template-rows: 40px 0px 100fr 0px;
 }
--- a/src/js/views/portals/PortalSectionView.js
+++ b/src/js/views/portals/PortalSectionView.js
@@ -144,6 +144,8 @@ define(["jquery",
               }
           });
 
+          document.body.style.removeProperty("height");
+
           this.markdownView?.tocView?.setAffix();
         },
 
--- a/src/js/views/portals/PortalView.js
+++ b/src/js/views/portals/PortalView.js
@@ -1060,6 +1060,14 @@ define(["jquery",
 
           MetacatUI.appModel.resetTitle();
           MetacatUI.appModel.resetDescription();
+
+          // Run subView onClose functions if they exist
+          for (const subView of this.subviews) {
+            if (typeof subView?.onClose === "function") {
+              subView.onClose();
+            }
+          }
+
           //Remove each subview from the DOM and remove listeners
           _.invoke(this.subviews, "remove");
 
--- a/src/js/views/portals/PortalVisualizationsView.js
+++ b/src/js/views/portals/PortalVisualizationsView.js
@@ -181,8 +181,23 @@ define(["jquery",
           this.$("iframe").css("height", remainingHeight + "px");
         },
 
+        /**
+         * Clean up
+         */
         onClose: function() {
-          $(window).removeListener("resize", this.adjustVizHeight);
+          // Remove body height style tag
+          document.body.style.removeProperty("height");
+
+          // this is failing for Cesium pages but might be necessary in FEVer ones
+          try {
+            $(window).removeListener("resize", this.adjustVizHeight);
+          }
+          catch (error) {
+            console.log(
+              'Failed to remove resize listener in portal viz onClose function. ' +
+              'Error details: ' + error
+            );
+          }
         }
 
      });
--- a/src/js/views/portals/PortalVisualizationsView.js
+++ b/src/js/views/portals/PortalVisualizationsView.js
@@ -137,6 +137,13 @@ define(["jquery",
               }
             }
   
+            if( this.model.get("visualizationType") == "cesium" ){
+              document.body.style.setProperty("height", "100%");
+            }
+            else {
+              document.body.style.removeProperty("height");
+            }
+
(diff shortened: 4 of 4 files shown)


robyngit (changes requested):
Hey @iannesbitt! Good fix, thanks for picking up this task and working on it so fast (and also finding it in the first place)! One issue I found is that the `height` property that gets set on cesium viz portal pages doesn't get removed if you navigate away from a portal. You can see this if you test locally:
1. Configure the theme to arctic and go to http://localhost:3000/portals/permafrost
2. See that `height: 100%` is set on `body`
3. Click on "Hosted by the Arctic Data Center" in the upper left hand corner
4. See that `height: 100%` is still set on `body`
5. Refresh the page and see that `height: 100%` is no longer on the body

Although this isn't causing issues for this particular page, it could conflict with layout styles on other pages, themes, or in the future. It's best to ensure that we always remove this property when we navigate away from the cesium portal page. Look into how `onClose` is used throughout metacatui, triggered by `AppView.showView`. This is one way to go about it, but definitely not the only way.

iannesbitt:
That makes total sense, good catch. I originally tried to put it in `PortalVisualizationsView.onClose()` but that function didn't seem to be called when I clicked on another portal section link, and I couldn't figure out why. I didn't think anything of it but I guess that's probably a larger issue to address as well...

robyngit:
@iannesbitt some background for you: `onClose` is a MetacatUI pattern, not a backbone thing (like I originally assumed when I started working on MetacatUI). Together, the router and AppView orchestrate removing and rendering the main views that make our single page application appear to navigate between different pages. `AppView` calls `onClose` on these "main" views before removing them. The idea is that `onClose` on these main views should then check for `onClose` methods on any sub-views and call those as well. Unfortunately, I think these `parent.onClose() --> child.onClose()` links are missing between views. We once were trying to establish a pattern of saving any `subviews` as a property of the parent view to help facilitate this, but this practice isn't followed consistently in the codebase.

iannesbitt:
@robyngit That makes sense given that PortalView has a working `onClose` function, but the PortalVisualizationsView one is never called. How are those links created?

robyngit:
> How are those links created?

@iannesbitt it's a matter of the parent views calling onClose on their subviews when they are closed. Here's an example of how that might look:

```js
ParentView = Backbone.View.extend({
  // ... other parent view methods ...

  createSubView: function () {
    const subView = new SubView();
    // When the parent view creates a subview, it should save a reference to it
    // so that it can call methods on it later.
    this.set("subView", subView);
    // or:
    // this.set('subViews', this.get('subViews').concat(subView));
  },

  onClose: function () {
    // When the parent view is closed, it should close the subview as well.
    const subView = this.get("subView");
    if (typeof subView?.onClose === "function") {
      subView.onClose();
    }
  },
});
```

So it involves a bit of manual work to make sure that the parent view is aware of its subviews and can call `onClose` on them when it is closed itself. There are likely less repetitive and less error-prone ways to implement this behaviour throughout MetacatUI, but this is the convention historically used here and maybe the quickest way to implement it without refactoring the entire codebase.

iannesbitt:
I get an error when trying `this.set(...)` in `PortalView.addVizSection()` and `PortalView.addSection()` that says the set function does not exist. So instead, I tried going with what's there. It looks like what happens is that the `sectionView` subview is appended to `this.subviews` using `push()`. I tried using `pop()` to get the last element in that array to grab the sectionView and run its `onClose` function, but I get `Uncaught TypeError: undefined has no properties`.

For comparison, when I add `document.body.style.removeProperty("height");` directly to `PortalView.onClose()`, everything works as expected.

This feels like an issue stemming from my lack of js experience, so my apologies if I'm missing something obvious...

robyngit:
No worries @iannesbitt, the issue you're experiencing is a common source of confusion and errors in JS! The value of the keyword `this` in JS can change depending on context, very unlike `self` in Python. Here is a good [short article](https://dev.to/ibtesum/this-in-javascript-and-self-in-python-any-difference-38fm) describing distinction between the two.

So I think you're getting the error saying that `set` does not exist because `this` isn't what you think it is. You may expect it to be the View that contains the method that is being called, but it may be the object that's calling the method or something else.  Check what `this` is in your code where you're calling `this.set` (with either a `console.log(this)` or a breakpoint in your browser's dev tools) and that might help you figure out what's going wrong.

If it helps, [`Function.prototype.call()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/call) can be used to set the value of `this` to a specific object within a function.

`document.body.style.removeProperty("height")` works because `document` is a global object and available everywhere, including in PortalView.

robyngit:
Works and looks great, @iannesbitt! Thank for finding and fixing this!!

robyngit (approved):

PR #2300, opened 2024-03-13
Fix return type jsdoc for Toolbar rendering section content

ianguerin:


robyngit (approved):
FYI you can also do:

```js
/**
 * @returns {{contentContainer: HTMLElement, sectionContent: Backbone.View}} The content container with the rendered view, and the Backbone.View itself.
 * - `contentContainer` is the content container HTML element.
 * - `sectionContent` is the Backbone.View instance.
 */
```

Issue #1795, opened 2021-05-19
Enable searching for specific data layers in Cesium


robyngit:
Some Cesium views could have a long list of (potentially nested) data layers. We should enable some sort of search box so that a user who is looking for a specific data layer can find it by either the title, description, or attribute

PR #2305, opened 2024-03-13
Feature 1795 layer search

yvonnesjy:
Add searching functionality to the layers panel- compatible with both "layers" view and "layerCategories" view. Update the view finder panel to share the search box component with the layers panel.

Fixes https://github.com/NCEAS/metacatui/issues/1795
Original PR: https://github.com/yvonnesjy/metacatui/pull/1


src/css/map-view.css
```diff
@@ -475,11 +479,85 @@ represents 1 unit of the given distance measurement. */
...
+/*****************************************************************************************
+ *
+ * Search Input
+ *
+ * Search Input is a shared component for searching information in the map toolbar.
+ *
+ */
+
+
+.search-input {
```
robyngit, line 491: @yvonnesjy I was going to remind you that we don't use any CSS preprocessors, but then saw that the rules were being applied and had to look this up! There is native CSS nesting now??? what??? 🎉🎉🎉 Thank you for showing me this!

src/js/views/maps/SearchInputView.js
```diff
@@ -0,0 +1,167 @@
...
+    * The HTML classes to use for this view's element
+    * @type {string}
+    */
+    className: "search-input",
+
+    /**
+     * The HTML classes to use for this view's HTML elements.
+     * @type {Object<string,string>}
+     */
+    classNames: {
```
robyngit, line 37: Curious why you made classNames an attribute of the view here, but a constant in other views?

src/css/map-view.css
yvonnesjy: ooooooh that's why nothing was nested! I guess all projects I worked on use preprocessors? cause I didn't know this was ever not an option haha. google says nesting was supported natively since last year🙂

src/js/views/maps/LayerCategoryItemView.js
```diff
@@ -103,13 +103,14 @@ define(
...
-          layerList.render();
-          this.$(`.${CLASS_NAMES.layers}`).append(layerList.el);
+          this.layerListView = new LayerListView({ collection: this.model.get('mapAssets') });
+          this.layerListView.render();
+          this.$(`.${CLASS_NAMES.layers}`).append(this.layerListView.el);
 
           // Show the category as expanded or collapsed depending on the model
           // properties.
           this.updateLayerList();
+          this.listenTo(this.model, 'change:expanded', this.updateLayerList);
```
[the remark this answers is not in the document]
yvonnesjy: all the updates within this view use direct DOM manipulation, no re-rendering. layer category comes from XML, so the content won't change either. no need to remove listener:)

src/css/map-view.css
yvonnesjy: ooooooh that's why nothing was nested! I guess all projects I worked on use preprocessors? cause I didn't know this was ever not an option haha. google says nesting was supported natively since last year🙂

robyngit: Amazing! Yes, before that, this was a syntax that was used in SCSS or Sass, and you would need to compile it to CSS for browser use. Perhaps MetacatUI will get away with never adopting a CSS preprocessor...

src/js/views/maps/SearchInputView.js
yvonnesjy: I def just copied the code from somewhere else haha. updated to constant since I fought for it🙂

yvonnesjy:
> I'm guessing it might be a case of overly specific CSS rules somewhere outside of the map view...

It's actually an issue with the input box background, using white from the default bootstrap config. I was not aware of testing on other portals when I sent out this PR- It should be fixed and working everywhere now:)

robyngit (approved):
Thanks for fixing @yvonnesjy! 🎉

PR #2314, opened 2024-03-19
Feature 1796 places autocomplete

ianguerin:
Includes all changes from individual PRs 1,2,3 and 5 (contents of 4 included in earlier PR) that were individually reviewed.

The goal of this PR is to be tested using a more manual review, by running in various themes.


robyngit (changes requested):
This is going to be fantastic @ianguerin, thanks for keeping it all so modular. I encountered some issues:
1. When there's no maps key configured in a repository, the ToolbarView breaks and the help tab view doesn't get rendered. The console shows the message: `Error rendering ToolbarView section TypeError: gmaps is null`
2. With a map key configured, I actually wasn't able to find any places. I only see the message: "No search results found."

ianguerin:
> This is going to be fantastic @ianguerin, thanks for keeping it all so modular. I encountered some issues:
> 
> 1. When there's no maps key configured in a repository, the ToolbarView breaks and the help tab view doesn't get rendered. The console shows the message: `Error rendering ToolbarView section TypeError: gmaps is null`
> 2. With a map key configured, I actually wasn't able to find any places. I only see the message: "No search results found."

I've hidden the viewfinder panel if there is not Google Maps API key now. 
I'm curious to reproduce #2 in your list, as it works with my personal API key, so I wonder if there's an API not enabled on the key

robyngit:
Tested with a properly configured API key, it works so well!! 🎉🎉🎉

src/js/models/geocoder/GoogleMapsAutocompleter.js
```diff
@@ -0,0 +1,47 @@
...
+      autocompleter = new gmaps.places.AutocompleteService();
+
+      /**
+       * Use the Google Maps Places API to get place predictions based off of a
+       * user input string as the user types.
+       * @param {string} input - User input to search for Google Maps places.
+       * @returns {Prediction[]} An array of places that could be the result the
+       * user is looking for. Most often this comes in five or less results.
+       */
+      async autocomplete(input) {
```
robyngit, line 25:
```suggestion
      async autocomplete(input) {
        if (!input) return [];
```

robyngit (approved):

Issue #2278, opened 2024-02-20
Update PDG font and color scheme


yvonnesjy:
Mock: https://www.sketch.com/s/eea6a4f3-5c59-40e6-bf7f-64edc511e4cb

Ideally apply to whole site. If color changes become tricky, panel only is acceptable

robyngit:
Colors can be configured in the portal document, as can a "layout" and a "theme". Let's implement these changes as a new "theme" such that it doesn't impact any of the existing portals.

yvonnesjy:
Documenting the following fixes that we will implement as a part of the April release (R1):

- Enable weights on Barlow font
- Remove text-transform treatments

PR #2317, opened 2024-03-21
Feature 2278 new portal theme

yvonnesjy:
Base PR: https://github.com/NCEAS/metacatui/pull/2305

This should only affect portals that adopt the new "light" theme. UI in existing portals (including PDG) should be noop. Changes in this PR have been reviewed by @ianguerin .

Mock: https://www.figma.com/file/9R80DBDAMQtOsu8y1HkAVT/permafrost-discovery-gateway?type=design&node-id=0-880&mode=design&t=Ex8HegJ018zpBftj-0

Fixes: https://github.com/NCEAS/metacatui/issues/2278

Testing setups:
- Add `modelJSON.theme = "light";` in [PortalModel.parse()](https://github.com/NCEAS/metacatui/blob/758ff8e3710637d9fdf406a9c2383e56ff5be877/src/js/models/portals/PortalModel.js#L522)
- To test layer categories, add lines from [layer categories config.txt](https://github.com/NCEAS/metacatui/files/14705125/layer.categories.config.txt) to [Map.initialize()](https://github.com/NCEAS/metacatui/blob/feature-2278-new-portal-theme/src/js/models/maps/Map.js#L234)


robyngit (changes requested):
Thank you for including a testing setup @yvonnesjy - that's super helpful! I'm seeing that setting the light theme is breaking other portal pages, see examples from PDG:

Screenshot 2024-03-21 at 4 55 39 PM
Screenshot 2024-03-21 at 4 55 44 PM
Screenshot 2024-03-21 at 4 55 59 PM
Screenshot 2024-03-21 at 4 56 05 PM

Given the... intricacy... of MetacatUI's CSS, it's not surprising this would happen. I was thinking that this "theme" is really more of a MAP theme than a portal theme. Is the theme intended to change the appearance of other parts of the portal? If not, I wonder how much work would it be to add a "theme" property to the map config and move the CSS changes to a map-light file (or something similar)? Maybe it would make it easier to apply the changes to the map specifically, rather than to the entire portal. Another very large added benefit is that the theme could be used on maps outside of portals, for example, in the main repository data catalog.

yvonnesjy:
Hey yup! I intentionally leave the other pages out for the following reasons:
- The mock only covers the map, so I don't want to arbitrarily assign colors to the other pages
- Incremental changes are good. We don't need to cram all page updates in one PR

RE: map theme- I'd vote against it and here's why:
- This PR doesn't update other pages, but it should def be a work item to do so in the future, when there is an agreed upon design. Outside of the in between state we're in right now, there is no reason to intentionally support a map theme with a different color palette than the portal theme.
- Supporting map in main repository is a use case I wasn't aware of, and that explains why ADC doesn't pick up the light theme! Though imo, having one map theme that applies to both main repository and portals is a recipe for chaos. What we actually need is for the map to understand where it is and pick up the theme accordingly. We can do so by adopting the same color palette schema in the repository metacatui.css (ex. --repo-col-text-body, etc.). Then in map-view.css, we can do `--map-col-text-body: var(--portal-col-text-body, var(--repo-col-text-body));`

yvonnesjy:
WAIT SORRY Robyn I didn't look at your screenshots closely and didn't test other tabs enough! I see that there's a background color change now and I'm on it!

robyngit:
> This PR doesn't update other pages, but for consistency, it should def be a work item to do so in the future, when there is an agreed upon design.

I see!! So the "light" theme WILL eventually be a portal theme 👍🏻 That makes sense.

> Supporting map in main repository is a use case I wasn't aware of

Yes! This is one of the cases that I keep testing for, because the MapView is used within the repository's data catalog search page (see the arctic version here: https://arcticdata.io/catalog). We have plans to integrate the Mapview in a few other places in MetacatUI as well, so it's important that it remains a standalone component, independent of the portals.

robyngit (dismissed):
🎉 looks fantastic!

robyngit (approved):
🎉 looks fantastic!

robyngit (dismissed):

robyngit (approved):

robyngit:
This PR seems to be a case of a GitHub bug that inappropriately dismisses PR approvals (see https://github.com/isaacs/github/issues/1157). The branch is approved and tests have passed so I am bypassing branch protections and merging the branch.

PR #2318, opened 2024-03-21
Feature 2277 layers list polish

yvonnesjy:
Base PR: https://github.com/NCEAS/metacatui/pull/2317

This PR updates the layers panel UI and adopts the new layer details button mechanism. The changes affect all portals and have been tested on PDG, ADC, DRP. Most changes besides the last commit have been reviewed by @ianguerin .

[Mock](https://www.figma.com/file/9R80DBDAMQtOsu8y1HkAVT/permafrost-discovery-gateway?type=design&node-id=0-880&mode=design&t=Ex8HegJ018zpBftj-0)
[Prototype](https://www.figma.com/proto/9R80DBDAMQtOsu8y1HkAVT/permafrost-discovery-gateway?page-id=60%3A41271&type=design&node-id=224-15217&viewport=143%2C173%2C1&t=0T3YWZspyuioCeg5-1&scaling=min-zoom&starting-point-node-id=70%3A47595)

Fixes: https://github.com/NCEAS/metacatui/issues/2278

Testing setups:

Add modelJSON.theme = "light"; in [PortalModel.parse()](https://github.com/NCEAS/metacatui/blob/758ff8e3710637d9fdf406a9c2383e56ff5be877/src/js/models/portals/PortalModel.js#L522)
To test layer categories, add lines from [layer categories config.txt](https://github.com/NCEAS/metacatui/files/14705125/layer.categories.config.txt) to [Map.initialize()](https://github.com/NCEAS/metacatui/blob/feature-2278-new-portal-theme/src/js/models/maps/Map.js#L234)


robyngit (changes requested):
This looks great!! I have two teeny tiny requests and a few comments

### Changes Requested

1. Could you please update the LayersPanelView screenshot to reflect the new layer list polish?
2. The eye icon is the same whether a layer is hidden or visible. Can we change it to the crossed out eye version when the layer is hidden?
Screenshot 2024-03-21 at 6 03 51 PM

### Comments

I have some ideas about potential spacing improvements, but maybe @hngrtr would be better at advising on this. When a layer item is two or more lines long, they appear crowded to me, and difficult to distinguish one item from the next. It seems to me that there should be more space between items and less space between lines of the same item:

Screenshot 2024-03-21 at 5 45 59 PM

Likewise, I think the list might look less crowded with just a little space at the bottom of each category's list, before the next category title:

spacing

Also the padding for the Viewfinder disappears with this PR, but I assume this is something that needs to be addressed once both features are merged into develop (just want to point this out, in case):

Screenshot 2024-03-21 at 5 13 13 PM

mbjones:
This all looks awesome. Thought I would throw in my 2 cents -- I noticed there is a size change on hover that happens when the Settings icon shows up when you hover on a row. This causes a visually disruptive shift in line height. Can we create a right justified margin for line length to leave room for the settings icon to show/hide on hover without shifting the line length? Here's a quick screen video of what I was seeing:


https://github.com/NCEAS/metacatui/assets/766407/d69c6564-b410-4048-b4da-66fb2c7d9bb9

hngrtr:
This is looking solid! Yvonne and I are meeting to go over outstanding
items and we also have the upcoming but bash. All of the sizing, spacing,
behavior, etc is in the Figma file for reference.


[image: Google Logo]
Doug Hungarter
Senior Staff UX Designer
Cloud Databases
***@***.***

yvonnesjy:
1. Eye icon: The new design uses color as a visibility indicator instead of the icon. I have a feeling it is an intentional choice. @hngrtr Can you chime in here?

2. Multi-line layer name: Similar to your suggestion, Matt- Doug and I synced and we will increase the toolbar width to accommodate, and also set a max-width on layer label + no wrapping allowed (overflow: ellipsis). That'll keep the labels in one line and hopefully alleviate the crowdedness.

3. Viewfinder: Yup I didn't touch it intentionally. Ian will have a followup PR to adopt the new styles.

yvonnesjy:
Added screenshots! @hngrtr Since we're in a race to get the PR's merged for bug bash, if you have the panel width and max-width for layer labels, I can incorporate them right now. Otherwise, I'm thinking of keeping the PR as is and covering the updates as the followup after tomorrow.

robyngit (approved):

PR #2322, opened 2024-03-25
Use shared SearchInputView in ViewfinderView

ianguerin:
Support additional callbacks within SearchInputView for keydown and keyup events

shared-input-field

### DIFF
--- a/src/css/map-view.css
+++ b/src/css/map-view.css
@@ -552,6 +552,7 @@ represents 1 unit of the given distance measurement. */
     .search-input__input {
       width: 100%;
       position: absolute;
+      padding-right: 32px;
     
       input[type=text]& {  /* Override bootstrap */
         height: 100%;
--- a/src/css/map-view.css
+++ b/src/css/map-view.css
@@ -1365,50 +1365,6 @@ other class: .ui-slider-range */
   width: 100%;
 }
 
-.viewfinder__field {
-  border-radius: 4px;
-  border: 1px solid var(--portal-col-bkg-active);
-  display: flex;
-  flex-direction: row;
-  background: var(--portal-col-bkg-lighter);
-
-  &:focus-within {
-    border-color: var(--map-col-input-error-border);
-    box-shadow: inset 0 1px 1px rgba(0, 0, 0, .075), 0 0 8px rgba(82, 168, 236, .6);
-  }
-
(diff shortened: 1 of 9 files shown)


src/js/views/maps/viewfinder/ViewfinderView.js
```diff
@@ -216,6 +184,34 @@ define(
...
 
+      /**
+       * Render the SearchInputView. 
+       */
+      renderSearchInput() {
+        this.searchInput = new SearchInputView({
+          placeholder: "Enter coordinates or areas of interest",
+          search: text => {
+            this.viewfinderModel.search(text);
+            // TODO(ianguerin): set error if results are empty.
```
yvonnesjy, line 195: Is this for future or a reminder for this PR?

ianguerin: No, just an out of date Todo, removing

src/css/map-view.css
```diff
@@ -1323,7 +1323,7 @@ other class: .ui-slider-range */
     border: 1px solid var(--portal-col-bkg-active);
     border-radius: 4px;
     box-sizing: border-box;
-    color: var(--map-col-text);
+    color: var(--map-col-text-body);
```
yvonnesjy, line 1326: --map-col-text-body is only defined when a theme is used, while some portals/main repos don't use themes (you can test on ADC or DRP). do we need a fallback color for those use cases?

same question for other color variables.

robyngit (changes requested):
Looks great!!! I added some suggestions where I thought we migth want to check that elements are defined before calling methods on them. As it is now, if one of the getSomeElement methods comes back undefined, then the rest of the Toolbar fails to render. Feel free to use a try/catch block or some other approach, just wanted to point that out.

ianguerin:
> Looks great!!! I added some suggestions where I thought we migth want to check that elements are defined before calling methods on them. As it is now, if one of the getSomeElement methods comes back undefined, then the rest of the Toolbar fails to render. Feel free to use a try/catch block or some other approach, just wanted to point that out.

jQuery selectors don't return undefined/null when there's no element, so I don't think that the optional chaining will have any effect. I will, however, add `|| ''` to the getInputValue function so that it always returns a string

robyngit (approved):

Issue #2277, opened 2024-02-20
Layers list polish


yvonnesjy:
- Move visibility toggle to left
- Change legend to settings button on hover
- Drop layers icon

Mock: https://www.sketch.com/s/eea6a4f3-5c59-40e6-bf7f-64edc511e4cb

yvonnesjy:
@robyngit Do we want to limit these changes to PDG? Or are we okay with applying them to all portals?

robyngit:
> Do we want to limit these changes to PDG? Or are we okay with applying them to all portals?

@yvonnesjy, I have a few questions myself before I can answer.

> Drop layers icon

In layer lists that are NOT grouped, will the icon be dropped or not? If we continue to show icons for non-grouped layer item, will the visibility toggle stay on the right?

> Change legend to settings button on hover

I don't understand this - the legend will only show when you hover over the layer settings button? Will it not work on touch devices then?



*(Reminder to update the documentation for the layer icon property of the map config if the icon will be deprecated or not show in some circumstances)*

hngrtr:
[The Figma prototype](https://www.figma.com/proto/9R80DBDAMQtOsu8y1HkAVT/permafrost-discovery-gateway?page-id=60%3A41271&type=design&node-id=70-47595&viewport=143%2C173%2C1&t=0T3YWZspyuioCeg5-1&scaling=min-zoom&starting-point-node-id=70%3A47595&mode=design) shows what I'm thinking here for desktop. Clicking a layer list item anywhere will toggle it's visibility, except for the gear icon that triggers the layer detail overlay. 

For mobile, I was thinking the primary tap action is toggling visibility like on desktop. The gear icon would shown when a layer is visible. 

For layers that aren't grouped, you could continue to use the icons that are there today, but have them double as visibility indicators. That way clicking a layer toggles visibility the same way whether or not the layers are grouped. You could use the light grey treatment that the PDG eye icon uses when a layer is off for the icons on a portal like DRP. When you click the DRP layer to turn it on, the icon would then turn solid and be blue (or whatever unique theme color the portal specifies).

As far as the legends go, the full versions are in the layer detail overlay. The swatches are shown when a layer is on more for a "palette overview" and to distinguish between the palettes being used in other enabled layers. Since some of these datasets (particularly categorial ones) can potentially have a lot of values in a legend, the interactive thumbnail won't scale up to support that.

[...]

Issue #1796, opened 2021-05-19
Include interface for searching for a place name or address in Cesium


robyngit:
Cesium includes an optional widget for searching for a location by address or name, and then navigating to the searched-for location. To enable this, we'll need a key for one of the [Geocoder services](https://cesium.com/docs/cesiumjs-ref-doc/?classFilter=GeoCoderService)

robyngit:
Also look into using the [GNIS database](https://stage.gnis-ld.org:30443/) for U.S. placenames, and [geonames](http://www.geonames.org/) for places outside the U.S.

robyngit:
We have a list of placenames that we are going to add as a layer to the PDG. It would be great to incorporate these into the places search, see https://github.com/PermafrostDiscoveryGateway/pdg-portal/issues/5

robyngit:
From #2229 (duplicate):

> PDG and other portals would be more usable if users could search for and zoom to specific places, such as named local communities, rivers, and other relevant places. This was discussed and raised at the PDG mini-workshop. Need to develop out the ideas here, this ticket is just a placeholder.

> One issue is we would need to be able to search place-name to location gazeteers, such as GNIS (see https://gnis-ld.org/ that we run, and https://geonames.org/).

ianguerin:
I've outlined a recommendation in [a Google doc exploring various geocoder APIs](https://docs.google.com/document/d/1crOvyK_ToQUbe7vRtcXS9-2KnmKhE90kLHATzioQzyg/). With input from Matt, Robyn, and Yvonne, we've elected to go with the Google Maps API. Below is a summary of the discussion had there in the case that this Google doc is no longer reachable.

There are a few reasons we've chosen this API. It supports international place name geocoding, it is an industry standard that we (MetacatUI) are already integrated with for using the Google Maps UI component, there's no additional server-side setup, and the story for integrating with the Places API for generating place name autocomplete suggestions similarly ought to be easy given the expected compatibility between Google APIs. 

In the discussion we also noted that the desire to search for local Arctic place names may become a challenge with Google API place names, so we should implement this integration in such a way that if we need to migrate to a different third-party service it is impossible or even very difficult.

mbjones:
[...]

PR #2324, opened 2024-03-26
Add margins & theme colors to viewfinder

robyngit:
Fixes display of viewfinder in the "light" portal theme

Screenshot 2024-03-26 at 11 15 28 AM
Screenshot 2024-03-26 at 11 06 40 AM

Related to issues #1796, #2277

### DIFF
--- a/src/css/map-view.css
+++ b/src/css/map-view.css
@@ -1365,15 +1365,19 @@ other class: .ui-slider-range */
   width: 100%;
 }
 
+.viewfinder__search-input {
+  margin: 0 24px;
+}
+
 .viewfinder-predictions {
   list-style: none;
-  margin: 0;
+  margin: -16px 24px 0 24px;
 
   .viewfinder-prediction__content {
     align-items: center;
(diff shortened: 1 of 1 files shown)


src/css/map-view.css
```diff
@@ -1365,15 +1365,19 @@ other class: .ui-slider-range */
...
 }
 
+.viewfinder__search-input {
+  margin: 0 24px;
+}
+
 .viewfinder-predictions {
   list-style: none;
-  margin: 0;
+  margin: -16px 24px 0 24px;
```
ianguerin, line 1374: Instead of negative margins, I was planning 

```
.viewfinder__predictions,
.viewfinder__search-input {
  margin: 0 24px;
}
```

ianguerin (approved):

src/css/map-view.css
robyngit: Yeah, that's what I did initially. The negative margin was an ugly fix for what I thought was too large of a gap between the input and the predictions list:

Screenshot 2024-03-26 at 1 08 44 PM

PR #2326, opened 2024-03-26
Fix styles in viewfinder

ianguerin:
This change:
- updates behaviors for shared search input to show a cancel button as soon as user types anything
- removes absolute positioning from input box to make space for a variable number of buttons in the input field
- always shows the search icon and gives it treatment when clicking it would trigger a search

after: 
someplace

before:
someplace--before

[search input changes.webm](https://github.com/NCEAS/metacatui/assets/4664309/e34f5701-2125-4d43-9111-d0a78c25aad5)

### DIFF
--- a/src/css/map-view.css
+++ b/src/css/map-view.css
@@ -548,11 +548,12 @@ represents 1 unit of the given distance measurement. */
     align-items: center;
     height: 40px;
     margin-bottom: 8px;
+    border: 1px solid;
+    border-color: var(--map-col-border-muted);
+    border-radius: 4px;
   
     .search-input__input {
       width: 100%;
-      position: absolute;
-      padding-right: 32px;
     
       input[type=text]& {  /* Override bootstrap */
         height: 100%;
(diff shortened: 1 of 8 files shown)


src/css/map-view.css
yvonnesjy, line 1: The background of the error state seems off: 
Screenshot 2024-03-27 at 12 00 56 AM

src/css/map-view.css
```diff
@@ -561,43 +562,68 @@ represents 1 unit of the given distance measurement. */
...
   
       .search-input__button-icon {
         font-size: 16px;
       }
     }
 
+    .search-input__vertical-divider {
+      width: 0;
+      height:24px;
+      border-right: solid 2px var(--map-col-border-muted);
```
yvonnesjy, line 611: I think the mock uses blue/2 instead?

yvonnesjy, line 611: should this be --map-col-border instead?

src/css/map-view.css
```diff
@@ -561,43 +562,68 @@ represents 1 unit of the given distance measurement. */
...
+    }
+
     .search-input__search-button {
       color: var(--map-col-buttons-icon-muted);
+      border-radius: 4px;
+
+      &.search-input__search-button--active {
+        color: var(--map-col-text-highlight, var(--map-col-text-body));
```
yvonnesjy, line 619: --map-col-text-body might not work as a fallback since it's from the new set of theme variables. fwiw currentColor has worked for me when I need text colors

yvonnesjy: instead of --map-col-text-highlight, can we use --map-col-buttons-icon?

src/css/map-view.css
```diff
@@ -561,43 +562,68 @@ represents 1 unit of the given distance measurement. */
...
+        background: var(--map-col-content-bkg-highlight, var(--map-col-bkg-lighter__deprecate));
```
yvonnesjy, line 620: instead of --map-col-content-bkg-highlight, can we create a new --...-col-buttons-bkg-active?

src/css/map-view.css
```diff
@@ -1376,12 +1402,10 @@ other class: .ui-slider-range */
   .viewfinder-prediction__content {
     align-items: center;
     background: var(--map-col-content-bkg, var(--map-col-bkg__deprecate));
-    /* TODO(ianguerin) */
-    border: 1px solid var(--map-col-border);
+    border: 1px solid var(--map-col-border, white);
```
yvonnesjy, line 1405: should this be --map-col-border-muted instead?

src/css/map-view.css
```diff
@@ -561,43 +562,68 @@ represents 1 unit of the given distance measurement. */
...
-        }
       }
     }
+
+    &:focus-within {
+      border-color: var(--map-col-border-highlight-search-box, var(--map-col-blue));
+    }
+
+    &.search-input__error-input,
+    &.search-input__error-input input[type=text].search-input__input {
```
yvonnesjy, line 579: is this selector repetitive?

src/css/map-view.css
yvonnesjy: nit: 1px instead of 2?

yvonnesjy (changes requested):

src/css/map-view.css
yvonnesjy, line 1: For viewfinder-prediction__focused, can we update border color to be --map-col-border-highlight?

yvonnesjy (changes requested):

src/css/map-view.css
ianguerin: Done

src/css/map-view.css
ianguerin: Done

src/css/map-view.css
ianguerin: I removed absolute positioning, so the input field and the parent div both need to be styled when there is an error. I'll move the styles to just change background of the parent div, and the input element will have no background.

src/css/map-view.css
ianguerin: Done

src/css/map-view.css
ianguerin: Done, done

src/css/map-view.css
```diff
@@ -1376,12 +1402,10 @@ other class: .ui-slider-range */
...
     align-items: center;
     background: var(--map-col-content-bkg, var(--map-col-bkg__deprecate));
-    /* TODO(ianguerin) */
-    border: 1px solid var(--map-col-border);
+    border: 1px solid var(--map-col-border, white);
     border-radius: 4px;
     box-sizing: border-box;
-    /* TODO(ianguerin) */
-    color: var(--map-col-text-body);
+    color: var(--map-col-text-highlight, var(--map-col-text-body));
```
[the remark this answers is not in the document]
ianguerin: I'm not sure the context of "ditto" here. Which other comment are you referring to?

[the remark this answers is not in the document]
ianguerin: Changing fallback to white

src/css/map-view.css
ianguerin: fixed-error-bkg
Fixed so that the whole input box is same color, the fallback for the orange is the lighter blue.

yvonnesjy (approved):

PR #2329, opened 2024-03-27
Allow navigating directly to a page in the Cesium-based data catalog

robyngit:
triggerSearch in Filters-Search connector was resetting the page to 1 because it was passing on arguments from the listener that called it. The argument was triggering the 'resetPage' option in the triggerSearch method.

### DIFF
--- a/src/js/models/connectors/Filters-Search.js
+++ b/src/js/models/connectors/Filters-Search.js
@@ -56,7 +56,7 @@ define([
         const search = this.get("searchResults");
 
         // Start results at first page and recreate query when the filters change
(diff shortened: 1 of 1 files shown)


iannesbitt (changes requested):
Tested and working on `arctic` and `drp`, but for some reason not working on `knb` or `dataone`. It seems to go to the selected page in the URL initially but then flips back right before the page load finishes.

iannesbitt:
Revising my review to an approve because the issue I saw is only applicable to themes that still use the google maps api (`knb` and `dataone`). We have decided to go ahead with this fix since that feature is slated for deprecation.

iannesbitt (approved):

PR #2330, opened 2024-03-27
Ensure page numbers are always >0 in catalog search

### LINKED ISSUE #2155: Clicking back button too many times makes start parameter negative
**Describe the bug**
When browsing DataONE search and using the browser back button to go back to the index page, hitting the back button too many times causes an error:

**To Reproduce**
1. Visit search.dataone.org
2. click on the a dataset
3. use the browser back button to go back to the index
4. Hit the back button again

**Expected behavior**
An error should not appear

**Screenshots**
See also https://app.screencast.com/U5XaleOOx1y5K

image

 - OS: OSX
 - Chrome
 - Version 114.0.5735.198 (Official Build) (x86_64)

robyngit:
Don't allow router to navigate to page numbers <0 in the catalog search

Fixes #2155

### DIFF
--- a/src/js/routers/router.js
+++ b/src/js/routers/router.js
@@ -299,8 +299,8 @@ function ($, _, Backbone) {
 			this.routeHistory.push("data");
 			// Check for a page URL parameter
 			page = parseInt(page);
-			page = isNaN(page) || page < 1 ? 0 : page;
-			MetacatUI.appModel.set("page", page);
+			page = (isNaN(page) || page < 1) ? 1 : page;
+			MetacatUI.appModel.set("page", (page-1));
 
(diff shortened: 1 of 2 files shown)


iannesbitt (changes requested):
This actually causes a perhaps inverse issue where clicking back after clicking on a dataset causes the page number being viewed in the browser to increment up by one. For instance, if the user clicks on a dataset on Page 3, they will be taken "back" to page 4 when they navigate back and so on.

robyngit:
Good catch @iannesbitt, thanks for testing so well! I think the latest commit fixes that problem 🤞🏻

iannesbitt (approved):
Tested again and working!

PR #2352, opened 2024-04-10
Fix issue with error not being cleared when user deletes and retypes queries

ianguerin:
[Various query combinations that previously caused issue.webm](https://github.com/NCEAS/metacatui/assets/4664309/f727dffd-431c-43c8-9915-591fb75202a1)


robyngit (approved):
🎉

yvonnesjy (approved):

PR #2353, opened 2024-04-10
Map Components Polish 1: Toolbar links tweaks

yvonnesjy:
Light before and after:

Screenshot 2024-04-10 at 4 49 37 PM 
image

Dark before and after:

image 
image

PR #2354, opened 2024-04-10
Map Components Polish 2: Add top shadow

yvonnesjy:
This adds a subtle shadow on top of the map to soften the transition from the top navigation bar to the map.

Before
image

After
image

### DIFF
--- a/src/css/map-view.css
+++ b/src/css/map-view.css
@@ -100,6 +100,11 @@
 
   /* FONTS */
   --map-body-font: var(--portal-body-font);
+
+  /* ORDERING */
+  --map-z-index-base: 1;
+  --map-z-index-top-shadow: calc(var(--map-z-index-base) + 1);
+  --map-z-index-layer-details: calc(var(--map-z-index-top-shadow) + 1);
 }
 
 /* hide the credits until we can find a better placement for them */
@@ -252,7 +257,7 @@
   height: 2rem;
   box-shadow: 0 0.2rem 0.2rem 0 rgba(0, 0, 0, .16) inset;
   width: 100%;
-  z-index: 2;
+  z-index: var(--map-z-index-top-shadow);
 }
 
 /* On larger screens, the toolbar should 'squish' the map, not overlap it */
@@ -290,7 +295,7 @@
   top: 1rem;
   right: 1rem;
   /* required to be placed above map widget in firefox: */
-  z-index: 1;
+  z-index: var(--map-z-index-base);
 }
 
 .map-view__toolbar-container {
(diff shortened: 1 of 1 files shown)


robyngit:
Tested this but then realized you've only set Ian as the reviewer, oopsie! Here is what I found:

- The issue in https://github.com/NCEAS/metacatui/pull/2353 is included in these changes.

- The help panel screenshot includes this new feedback blurb, which isn't actually included in these changes: Screenshot 2024-04-11 at 10 55 34 AM
  (BTW, where will we source the contact email from when we do decide to implement this feedback blurb?)

yvonnesjy:
> The issue in https://github.com/NCEAS/metacatui/pull/2353 is included in these changes.

Hey yup, this branch is based on the 2353 branch (and all my styling changes are kinda based on each other now). In general, I name my _initial_ commit for a new PR `<issue name>: <change summary>` and the followup commits `Fixup: <change summary>`. So when you're reviewing, you can select the _last_ commit that _doesn't_ start with `Fixup` and all the fixups that follow it.

> The help panel screenshot includes this new feedback blurb

That was from https://github.com/NCEAS/metacatui/pull/2348/files and the screenshot **with** the feedback blurb actually existed in the repo, and was replaced by one **without** the blurb. I just thought it was an outdated screenshot cause I haven't seen blurb anywhere... but lmk if that's not the case😳

ianguerin (approved):

robyngit (changes requested):
Reminder:
[ ] Ensure that the expanded layer panel doesn't overlap with the map in the Data Catalog View.

yvonnesjy:
> Ensure that the expanded layer panel doesn't overlap with the map in the Data Catalog View. 

Rebased and this should be fixed now.

src/css/map-view.css
```diff
@@ -339,7 +339,7 @@
   max-height: 100%;
   overflow: auto;
   /* required to be placed above map widget in firefox: */
-  z-index: 1;
+  z-index: 3;
```
ianguerin, line 342: In my experience this is the kind of thing that is destined to be a regression (especially without screenshot tests). One small thing that has worked for me is defining a variable for the z-index you're interested in, and then doing some math based on that z-index. So if you have 
```
--map-widget-z-index: 1; 
```

Then here you could have
```
z-index: calc(2 + var(--map-widget-z-index);
```

ianguerin (approved):

src/css/map-view.css
robyngit: ^ good tip @ianguerin, I'm going to use that

yvonnesjy: much better!

Issue #2275, opened 2024-02-20
Support "Zoom to..." predefined areas


yvonnesjy:
Fly to named areas (e.g. Utqiagvik, Southern Alaska) defined in portal XML config

Mock: https://www.sketch.com/s/eea6a4f3-5c59-40e6-bf7f-64edc511e4cb/a/K0me5W1

PR #2356, opened 2024-04-11
Add expansion panel view for showing/hiding content

ianguerin:
This component can be used to coordinate multiple expansion panels such that only one panel is ever open. 
This will be used in the viewfinder view to hide the zoom-to-presets UI.

[expansion panel component.webm](https://github.com/NCEAS/metacatui/assets/4664309/340ef5e2-d5f4-49ae-94a2-d63c0f0b3ef3)

[NCEAS/2275](https://github.com/NCEAS/metacatui/issues/2275)


yvonnesjy:
> coordinate multiple expansion panels such that only one panel is ever open

Is that UX design intent? It isn't clear in the mock or prototype, but I'm gathering that this mechanism isn't implemented in the prototype.

Why this matters: I'm wondering how much html/js/css code this component can share with layer category item. Layer categories def don't want to auto collapse others.

ianguerin:
> > coordinate multiple expansion panels such that only one panel is ever open
> 
> Is that the UX design intent? It isn't clear in the mock or prototype, but I'm gathering that this mechanism isn't implemented in the prototype.
> 
> Why this matters: I'm wondering how much html/js/css code this component can share with layer category item. Layer categories def don't want to auto collapse others.

Hmm, I swear in the prototype it used to close when another was opened. It looks like multiple can be opened at the same time in the prototype. I'll remove that. This component should be shareable

src/css/map-view.css
```diff
@@ -1410,3 +1410,53 @@ other class: .ui-slider-range */
     }
   }
 }
+
+/** START - Expansion panel View. */
+.expansion-panel {
+	border: solid var(--portal-col-bkg-active);
```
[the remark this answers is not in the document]
yvonnesjy: should this be --map-col-item-divider instead? and we probably need a fallback as well?

src/js/models/maps/viewfinder/ExpansionPanelsModel.js
```diff
@@ -0,0 +1,60 @@
...
+  * ExpansionPanelView instances so that only one is open at a time.
+  * @classcategory Models/Maps
+  */
+  const ExpansionPanelsModel = Backbone.Model.extend({
+    /**
+     * @name ExpansionPanelsModel#defaults
+     * @type {Object}
+     * @property {ExpansionPanelView[]} panels The expansion panel views that
+     * are meant to have only a single panel open at a time.
+     * @property {string} mode The open mode which determines the behavior of
```
yvonnesjy, line 16: would it be less typo prone to use a boolean instead? do we expect more than two modes?

src/js/templates/maps/viewfinder/expansion-panel.html
```diff
@@ -0,0 +1,14 @@
+<div class="<%= classNames.toggle %>">
+  <div class="<%= classNames.icon %>">
+    <i class="icon icon-<%= icon %>"></i>
```
yvonnesjy, line 3: there is a naming difference among different versions of fontawesome. the newer versions stopped using `icon-`. let's have the users specify the full class name to help with the migration here.

src/css/map-view.css
```diff
@@ -1410,3 +1410,53 @@ other class: .ui-slider-range */
     }
   }
 }
+
```
yvonnesjy, line 1413: Compared to layer-category-item, expansion-panel is missing some font/icon size and color treatments. are we saving the polish for later?

yvonnesjy (changes requested):

src/css/map-view.css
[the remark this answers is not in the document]
ianguerin: Somehow white spaces were tabs instead of white space characters; let's let a linter worry about that stuff in the future. Adding fallback CSS variables and using map variable.

src/js/templates/maps/viewfinder/expansion-panel.html
ianguerin: Done

src/js/models/maps/viewfinder/ExpansionPanelsModel.js
ianguerin: Done

src/css/map-view.css
```diff
@@ -1413,50 +1413,50 @@ other class: .ui-slider-range */
...
-		align-items: center;
-		display: flex;
-	}
-
-	.expansion-panel__toggle {
-		display: flex;
-		justify-content: space-between;
-		cursor: pointer;
-	} 
+  border: solid var(--map-col-item-divider, white);
```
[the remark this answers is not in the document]
ianguerin: Done

src/js/models/maps/viewfinder/ExpansionPanelsModel.js
```diff
@@ -0,0 +1,59 @@
+'use strict';
+
+define([], () => {
+  /**
+  * @class ExpansionPanelsModel
+  * @classdes ExpansionPanelsModel maintains state for multiple
```
robyngit, line 6:
```suggestion
  * @classdesc ExpansionPanelsModel maintains state for multiple
```

src/js/models/maps/viewfinder/ExpansionPanelsModel.js
```diff
@@ -0,0 +1,59 @@
...
+  * ExpansionPanelView instances so that only one is open at a time.
+  * @classcategory Models/Maps
+  */
+  const ExpansionPanelsModel = Backbone.Model.extend({
+    /**
```
robyngit, line 11:
```suggestion
  const ExpansionPanelsModel = Backbone.Model.extend(
    /** @lends ExpansionPanelsModel.prototype */{
    /**
```

src/js/models/maps/viewfinder/ExpansionPanelsModel.js
```diff
@@ -0,0 +1,59 @@
...
+     * @name ExpansionPanelsModel#defaults
+     * @type {Object}
+     * @property {ExpansionPanelView[]} panels The expansion panel views that
+     * are meant to have only a single panel open at a time.
+     * @property {boolean} isMulti Whether multiple panels can be open at the
+     * same time when displayed in a group of panels.
```
robyngit, line 17:
```suggestion
     * same time when displayed in a group of panels. 
     * @extends Backbone.Model
```

src/js/views/maps/viewfinder/ExpansionPanelView.js
```diff
@@ -0,0 +1,127 @@
...
+     * @classdesc Allow expand and collapse content in a panel.
+     * @classcategory Views/Maps/Viewfinder
+     * @name ExpansionPanelView
+     * @extends Backbone.View
+     * @screenshot views/maps/viewfinder/ExpansionPanelView_closed.png
+     * @screenshot views/maps/viewfinder/ExpansionPanelView_open.png
+     * @since x.x.x
+     * @constructs ExpansionPanelView
+     */
+    var ExpansionPanelView = Backbone.View.extend({
```
robyngit, line 24:
```suggestion
    var ExpansionPanelView = Backbone.View.extend(
      /** @lends ExpansionPanelView.prototype */{
```

src/js/views/maps/viewfinder/ExpansionPanelView.js
```diff
@@ -0,0 +1,127 @@
...
+    const BASE_CLASS = 'expansion-panel';
+
+    /**
+     * @class ExpansionPanelView
+     * @classdesc Allow expand and collapse content in a panel.
+     * @classcategory Views/Maps/Viewfinder
+     * @name ExpansionPanelView
+     * @extends Backbone.View
+     * @screenshot views/maps/viewfinder/ExpansionPanelView_closed.png
+     * @screenshot views/maps/viewfinder/ExpansionPanelView_open.png
```
robyngit, line 20: Only one of these screenshots will show up in the docs. Here's a single image you could use:

ExpansionPanelView

src/js/models/maps/viewfinder/ExpansionPanelsModel.js
```diff
@@ -0,0 +1,59 @@
...
+        panels: [],
+        isMulti: false,
+      }
+    },
+
+    /**
+     * @param {boolean} isMulti The display mode of the panels.
+     */
+    initialize({ isMulti } = { isMulti: false }) {
+      this.set('isMulti', isMulti);
```
robyngit, line 30: attributes passed to the model will [automatically be set](https://backbonejs.org/#Model-constructor). Is this needed? Is it intentional to prevent any additional attributes from being set (e.g. panels)?

src/js/views/maps/viewfinder/ExpansionPanelView.js
```diff
@@ -0,0 +1,127 @@
...
+      /**
+       * The type of View this is
+       * @type {string}
+       */
+      type: 'ExpansionPanelView',
+
+      /**
+       * The HTML class to use for this view's outermost element.
+       * @type {string}
+       */
```
robyngit, line 34: I think because we have `@extends Backbone.View`, you can simply use `@inheritdoc` for any methods are part of the base backbone view. I realize we're really not making use of this in older views in MetacatUI, but thought you should know about that shortcut.

robyngit (changes requested):
Hurray for reusable components! 🎉 Wondering if this could/should be a view that we use throughout MetacatUI, outside of just the maps?

src/js/models/maps/viewfinder/ExpansionPanelsModel.js
ianguerin: Nope, if its not necessary, I'll remove.

src/js/views/maps/viewfinder/ExpansionPanelView.js
ianguerin: Done

ianguerin:
We'll need to move it if we want to adopt it in the layer search UI as well, I think we can just plan to move it to a higher common folder if the need arises elsewhere. Same goes for our shared search input as well

PR #2377, opened 2024-04-19
Adding colors to dark theme for #2376

### LINKED ISSUE #2376: Colors in portal dark layer list don't quite fit after update
**Describe the bug**
The dark theme needs to be refreshed after the updates to the light theme from 2.28.0. Dark theme colors are sometimes too dark to distinguish from the background.

**To Reproduce**
1. Go to https://drp.dataone.org/portals/drp-data
2. Look at layer list
3. See issue

**Expected behavior**
Theme colors should be readable.

**Screenshots**
Screen Shot 2024-04-19 at 15 57 46

 - OS: all
 - Browser: all
 - Version: all

**Smartphone (please complete the following information):**
 - Device: all
 - OS: all
 - Browser: all
 - Version: all

**Additional context**
n/a

iannesbitt:
Closes #2376

### DIFF
--- a/src/css/portal-themes/dark.css
+++ b/src/css/portal-themes/dark.css
@@ -1,5 +1,34 @@
 :root {
   /* COLOURS */
+  --portal-blue-1: #ECF8FC;
+  --portal-blue-2: #CCF0FC;
+  --portal-blue-3: #99E1FA;
+  --portal-blue-4: #14c8ff;
+  --portal-blue-5: #00A9E4;
+  --portal-blue-6: #0087B5;
+  --portal-blue-7: #136682;
+
+  --portal-grey-1: #F8F9FA;
+  --portal-grey-2: #F1F3F4;
+  --portal-grey-3: #E8EAED;
+  --portal-grey-4: #BDC1C6;
+  --portal-grey-5: #80868B;
+  --portal-grey-6: #5F6368;
+  --portal-grey-7: #3C4043;
+  --portal-grey-8: #202124;
+
+  --portal-col-buttons-bkg: white;
+  --portal-col-buttons-bkg-hover: var(--portal-blue-2);
+  --portal-col-buttons-bkg-highlight: var(--portal-blue-5);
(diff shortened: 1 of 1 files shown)


src/css/portal-themes/dark.css
```diff
@@ -1,5 +1,34 @@
...
+  --portal-blue-7: #136682;
+
+  --portal-grey-1: #F8F9FA;
+  --portal-grey-2: #F1F3F4;
+  --portal-grey-3: #E8EAED;
+  --portal-grey-4: #BDC1C6;
+  --portal-grey-5: #80868B;
+  --portal-grey-6: #5F6368;
+  --portal-grey-7: #3C4043;
+  --portal-grey-8: #202124;
```
yvonnesjy, line 18: Hi Ian! This color palette is specific to the light theme so I'd recommend not directly copying over the config from light.css. The issue described in the ticket is about layer category color, which is defined [here](https://github.com/NCEAS/metacatui/blob/af0cbddb2d22f5c1c5dbe4f6fd9867a207fec8bb/src/css/map-view.css#L859-L866). I think the fix should be two stepped:

1. Set --portal-col-text-highlight in dark.css to a legible color in the context of dark theme. This variable is used by other components as well (.layer-details__label, .viewfinder-prediction__content) so you'll see a color change at those places. This fix will also only apply to portals that explicitly use the dark theme. Map in the main repo, for example, will still use the fallback color of --map-col-highlight__deprecate.

2. To address the issue for non-theme users, we can replace the fallback color in map-view.css with the same color from # 1.

yvonnesjy (changes requested):

src/css/portal-themes/dark.css
iannesbitt: Hi @yvonnesjy! Thank you, I appreciate the heads up and the detailed explanation. I just assumed that copying the blues and greys definitions to dark.css and changing the specific variable references would be sufficient to make the dark theme more legible. I can change to define them directly.

Issue #2361, opened 2024-04-15
Modernize map pins for vector data


yvonnesjy:
Current: <img width="37" alt="image" src="https://github.com/NCEAS/metacatui/assets/8570665/5dfeb08d-ceaf-4810-b55a-ae278b6e58f1">

Design mock: <img width="37" alt="image" src="https://github.com/NCEAS/metacatui/assets/8570665/43bf1885-f3ff-4268-89e0-01ee2269ac89">

PR #2378, opened 2024-04-19
Update Pin Style

yvonnesjy:
Because Cesium Billboard only accepts Image or Canvas, we have to do the javascript svg -> image conversion. No good way to use `<i />` or style in map-view.css.

image -> image

Fixes: #2361

### DIFF
--- a/src/js/models/maps/assets/CesiumVectorData.js
+++ b/src/js/models/maps/assets/CesiumVectorData.js
@@ -16,7 +16,8 @@ define([
   VectorFilters
 ) {
   // Source: https://fontawesome.com/v6/icons/location-dot?f=classic&s=solid
-  const PIN_SVG_STRING = "<svg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 384 512'><!--!Font Awesome Free 6.5.2 by @fontawesome - https://fontawesome.com License - https://fontawesome.com/license/free Copyright 2024 Fonticons, Inc.--><path d='M215.7 499.2C267 435 384 279.4 384 192C384 86 298 0 192 0S0 86 0 192c0 87.4 117 243 168.3 307.2c12.3 15.3 35.1 15.3 47.4 0zM192 128a64 64 0 1 1 0 128 64 64 0 1 1 0-128z'/></svg>";
+  // !Font Awesome Free 6.5.2 by @fontawesome - https://fontawesome.com License - https://fontawesome.com/license/free Copyright 2024 Fonticons, Inc.
+  const PIN_SVG_STRING = "<svg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 384 512'><path d='M215.7 499.2C267 435 384 279.4 384 192C384 86 298 0 192 0S0 86 0 192c0 87.4 117 243 168.3 307.2c12.3 15.3 35.1 15.3 47.4 0zM192 128a64 64 0 1 1 0 128 64 64 0 1 1 0-128z'/></svg>";
   const B64_START = 'data:image/svg+xml;base64,';
 
   /**
--- a/src/js/models/maps/assets/CesiumVectorData.js
+++ b/src/js/models/maps/assets/CesiumVectorData.js
@@ -16,7 +16,7 @@ define([
   VectorFilters
 ) {
   // Source: https://fontawesome.com/v6/icons/location-dot?f=classic&s=solid
-  const PIN_SVG_STRING = "<svg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 384 512'><!--!Font Awesome Free 6.5.2 by @fontawesome - https://fontawesome.com License - https://fontawesome.com/license/free Copyright 2024 Fonticons, Inc.--><path d='M215.7 499.2C267 435 384 279.4 384 192C384 86 298 0 192 0S0 86 0 192c0 87.4 117 243 168.3 307.2c12.3 15.3 35.1 15.3 47.4 0zM192 128a64 64 0 1 1 0 128 64 64 0 1 1 0-128z'></svg>";
+  const PIN_SVG_STRING = "<svg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 384 512'><!--!Font Awesome Free 6.5.2 by @fontawesome - https://fontawesome.com License - https://fontawesome.com/license/free Copyright 2024 Fonticons, Inc.--><path d='M215.7 499.2C267 435 384 279.4 384 192C384 86 298 0 192 0S0 86 0 192c0 87.4 117 243 168.3 307.2c12.3 15.3 35.1 15.3 47.4 0zM192 128a64 64 0 1 1 0 128 64 64 0 1 1 0-128z'/></svg>";
   const B64_START = 'data:image/svg+xml;base64,';
 
   /**
@@ -610,6 +610,9 @@ define([
         const pin = new DOMParser()
             .parseFromString(PIN_SVG_STRING, "image/svg+xml")
             .querySelector("svg");
+        pin.setAttribute("width", styles.markerSize);
+        pin.setAttribute("height", styles.markerSize);
+
         const pinPath = pin.querySelector("path");
         pinPath.setAttribute("fill", styles.color.toCssHexString());
         pinPath.setAttribute("fill-rule", "evenodd");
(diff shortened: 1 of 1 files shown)


src/js/models/maps/assets/CesiumVectorData.js
```diff
@@ -603,12 +606,20 @@ define([
        * @since 2.25.0
        */
       styleBillboard: function (entity, styles) {
-        if (!this.pinBuilder) {
-          this.pinBuilder = new Cesium.PinBuilder();
-        }
-        entity.billboard.image = this.pinBuilder
-          .fromColor(styles.color, styles.markerSize)
-          .toDataURL();
+        const pin = new DOMParser()
```
ianguerin, line 609: I'm no SVG expert, but could you just set these attributes on the string defined on line 18?

```
const PIN_SVG_STRING = "<svg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 384 512'><!--!Font Awesome Free 6.5.2 by @fontawesome - https://fontawesome.com License - https://fontawesome.com/license/free Copyright 2024 Fonticons, Inc.--><path d='M215.7 499.2C267 435 384 279.4 384 192C384 86 298 0 192 0S0 86 0 192c0 87.4 117 243 168.3 307.2c12.3 15.3 35.1 15.3 47.4 0zM192 128a64 64 0 1 1 0 128 64 64 0 1 1 0-128z'
stroke='white' fill-rule='evenodd'></svg>";
```

As I type out that example, I realize that some of these attributes are dynamic, so maybe can we just move up the ones that are hardcoded?

ianguerin (approved):

src/js/models/maps/assets/CesiumVectorData.js
yvonnesjy: The svg string was copied straight from fontawesome, I think there's benefit to keep our customization separate from the copied string. I'll add a comment to indicate where the string is from.

robyngit:
I'm excited to see this! The pins aren't working on my end. In the production PDG portal, if I show the Permafrost Tunnel Tour layer, the pin doesn't show up and the console gives me:

```
Failed to update CesiumVectorData model styles. TypeError: pin is null
```

It's the same for the Local News Stories and Communities layers.

Sometimes on the first try the old PIN in the wrong colour will show up! It's weirdly in consistent:

Screenshot 2024-04-22 at 11 08 11 AM


I'm pleased that the error is caught and the entire map doesn't crash though! :)

yvonnesjy:
Bizarrrre. I can't reproduce on my side. @robyngit Does the error come from [this line](https://github.com/NCEAS/metacatui/blob/113c4eeb1dd20822f78045c9df43c848bdfbd3da/src/js/models/maps/assets/CesiumVectorData.js#L610)? When the old pin shows up, do we get the same error?

I wonder if it's something with DOMParser.

robyngit:
@yvonnesjy seems like it works for me on Chrome and not on Firefox!

yvonnesjy:
Ahhh thanks for testing and catching it in FF🕵🏻‍♀️! Fixed!

yvonnesjy:
Rebased and pushed!

src/js/models/maps/assets/CesiumVectorData.js
```diff
@@ -15,6 +15,10 @@ define([
   AssetColorPalette,
   VectorFilters
 ) {
+  // Source: https://fontawesome.com/v6/icons/location-dot?f=classic&s=solid
+  const PIN_SVG_STRING = "<svg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 384 512'><!--!Font Awesome Free 6.5.2 by @fontawesome - https://fontawesome.com License - https://fontawesome.com/license/free Copyright 2024 Fonticons, Inc.--><path d='M215.7 499.2C267 435 384 279.4 384 192C384 86 298 0 192 0S0 86 0 192c0 87.4 117 243 168.3 307.2c12.3 15.3 35.1 15.3 47.4 0zM192 128a64 64 0 1 1 0 128 64 64 0 1 1 0-128z'/></svg>";
```
robyngit, line 19: I'm not sure if it would have an impact on rendering, but I wonder if we should/could remove the `<!--!Font Awesome Free 6.5.2 by @fontawesome - https://fontawesome.com License - https://fontawesome.com/license/free Copyright 2024 Fonticons, Inc.-->` part from the SVG string, and instead include it as a comment in the JS above? If we render 10,000s of pins, would it make a difference?

src/js/models/maps/assets/CesiumVectorData.js
```diff
@@ -603,12 +607,21 @@ define([
...
+            .parseFromString(PIN_SVG_STRING, "image/svg+xml")
+            .querySelector("svg");
+        pin.setAttribute("width", styles.markerSize);
+        pin.setAttribute("height", styles.markerSize);
+
+        const pinPath = pin.querySelector("path");
+        pinPath.setAttribute("fill", styles.color.toCssHexString());
+        pinPath.setAttribute("fill-rule", "evenodd");
+        pinPath.setAttribute("stroke", "white");
+        pinPath.setAttribute("stroke-width", styles.markerSize);
```
robyngit, line 620: Why does stroke width == marker size?

src/js/models/maps/assets/CesiumVectorData.js
```diff
@@ -603,12 +607,21 @@ define([
...
+        const pin = new DOMParser()
+            .parseFromString(PIN_SVG_STRING, "image/svg+xml")
+            .querySelector("svg");
+        pin.setAttribute("width", styles.markerSize);
+        pin.setAttribute("height", styles.markerSize);
+
+        const pinPath = pin.querySelector("path");
+        pinPath.setAttribute("fill", styles.color.toCssHexString());
+        pinPath.setAttribute("fill-rule", "evenodd");
+        pinPath.setAttribute("stroke", "white");
```
robyngit, line 619: Could stroke color be `styles.outlineColor.toCssHexString()` so that it's customizable? 

Similarly, should we respect the `styles.outline` property and make the stroke transparent if it is set to `false`?

robyngit (approved):
@yvonnesjy Approved with light suggestions for you to accept or reject as needed

src/js/models/maps/assets/CesiumVectorData.js
yvonnesjy: Not sure how much latency it'd cause, but I did confirm the comment results in a larger encoded string on line 623. Moved to comment above

src/js/models/maps/assets/CesiumVectorData.js
yvonnesjy: Doug handed this config to me so I'm not fully sure how it works. But here is a resulting svg string:

`<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 384 512" width="24" height="24">\x3C!--!Font Awesome Free 6.5.2 by @fontawesome - https://fontawesome.com License - https://fontawesome.com/license/free Copyright 2024 Fonticons, Inc.--><path d="M215.7 499.2C267 435 384 279.4 384 192C384 86 298 0 192 0S0 86 0 192c0 87.4 117 243 168.3 307.2c12.3 15.3 35.1 15.3 47.4 0zM192 128a64 64 0 1 1 0 128 64 64 0 1 1 0-128z" fill="#d1134c" fill-rule="evenodd" stroke="white" stroke-width="24"/></svg>`

You can see the outline only starts to show around width 10, and gets thicker when you increase the stroke width

src/js/models/maps/assets/CesiumVectorData.js
yvonnesjy: looks like outlineColor is used for polygon + point, and besides styling, is used to indicate selected vs not selected. the use for the white stroke here is really only for styling purposes- the [free svg](https://fontawesome.com/v6/icons/location-dot?f=classic&s=solid) doesn't have an outline but the [mock](https://www.figma.com/file/9R80DBDAMQtOsu8y1HkAVT/permafrost-discovery-gateway?type=design&node-id=0-3768&mode=design&t=tiKPvHDDeatYKtDr-4) does. idt users need that granular of a control for the pin styling, and I also don't want to opt in to the outline treatment on select. I think hardcode is sufficient here.

robyngit: Really good points @yvonnesjy !

PR #2380, opened 2024-04-20
Base Map Polish: Support saturation for CesiumImagery

yvonnesjy:
This is so that we can set OpenStreetMaps and the upcoming OpenTopoMap to greyscale.

Demo:
image -> image

### DIFF
--- a/src/js/models/maps/assets/MapAsset.js
+++ b/src/js/models/maps/assets/MapAsset.js
@@ -60,6 +60,9 @@ define([
        * opacity of the layer on the map, with 0 representing fully transparent and 1
        * representing fully opaque. This applies to raster (imagery) and vector assets,
        * not to terrain assets.
+       * @property {Number} [saturation = 1] A number between 0 and 1 indicating the
+       * saturation of the layer on the map, with 0 representing grayscale and 1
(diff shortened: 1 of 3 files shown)


ianguerin (approved):

src/js/models/maps/assets/MapAsset.js
```diff
@@ -142,6 +146,9 @@ define([
        * opacity of the layer on the map, with 0 representing fully transparent and 1
        * representing fully opaque. This applies to raster (imagery) and vector assets,
        * not to terrain assets.
+       * @property {Number} [saturation=1] - A number between 0 and 1 indicating the
```
robyngit, line 149: You can also set it to any number > 1 to increase saturation beyond the image's current level

robyngit (approved):

src/js/models/maps/assets/MapAsset.js
yvonnesjy: Indeed! I also realized negative number works as well, though it's not what I expected... I just copied over Cesium's documentation. https://cesium.com/learn/ion-sdk/ref-doc/ImageryLayer.html#saturation

PR #2383, opened 2024-04-22
Move search UI into its own view that is rendered by ViewfinderView

ianguerin:
This is in preparation of adding more sub views into the ViewfinderView


src/js/views/maps/viewfinder/SearchView.js
```diff
@@ -0,0 +1,234 @@
...
+     * This view requires a Google Maps API key in order to function properly,
+     * and must have the Geocoding API and Places API enabled.
+     * @classcategory Views/Maps
+     * @name SearchView
+     * @extends Backbone.View
+     * @screenshot views/maps/viewfinder/SearchView.png
+     * @since x.x.x
+     * @constructs SearchView
+     */
+    var SearchView = Backbone.View.extend({
```
robyngit, line 42:
```suggestion
    var SearchView = Backbone.View.extend(
    /** @lends SearchView.prototype */ {
```

src/js/views/maps/viewfinder/SearchView.js
```diff
@@ -0,0 +1,234 @@
...
+       * Values meant to be used by the rendered HTML template.
+       */
+      templateVars: {
+        classNames: CLASS_NAMES,
+      },
+
+      /**
+       * @typedef {Object} SearchViewOptions
+       * @property {Map} The Map model associated with this view allowing control
+       * of panning to different locations on the map.
```
robyngit, line 62: Map --> ViewfinderModel?

src/js/views/maps/viewfinder/SearchView.js
```diff
@@ -0,0 +1,234 @@
...
+      templateVars: {
+        classNames: CLASS_NAMES,
+      },
+
+      /**
+       * @typedef {Object} SearchViewOptions
+       * @property {Map} The Map model associated with this view allowing control
+       * of panning to different locations on the map. 
+       */
+      initialize({ viewfinderModel }) {
```
robyngit, line 64: Can/should we use a new ViewfinderModel if one isn't passed to the view?

src/js/views/maps/viewfinder/ViewfinderView.js
```diff
@@ -69,158 +65,22 @@ define(
...
       },
 
-      /**
-       * Getter function for the list of predictions. 
-       * @return {HTMLUListElement} Returns the predictions unordered list
-       * HTML element.
-       */
-      getList() {
-        return this.$el.find(`.${this.classNames.predictions}`);
+      getSearch() {
```
robyngit, line 72:
```suggestion
      /**
       * Get the SearchView element.
       * @returns {JQuery} The SearchView element.
       */
      getSearch() {
```

robyngit (changes requested):
Lovely! Just some minor requests

src/js/views/maps/viewfinder/SearchView.js
ianguerin: Done

src/js/views/maps/viewfinder/SearchView.js
ianguerin: I don't think it would be particularly useful, since the ViewfinderModel needs to have the Map model that corresponds to the Cesium application.  What do you think?

src/js/views/maps/viewfinder/SearchView.js
```diff
@@ -0,0 +1,236 @@
...
+    var SearchView = Backbone.View.extend(
+    /** @lends SearchView.prototype */ {
+      /**
+       * The type of View this is
+       * @type {string}
+       */
+      type: 'SearchView',
+
+      /** @inheritdoc */
+      className: BASE_CLASS,
```
yvonnesjy, line 51: is it intended? we probably don't need a class name here right?

yvonnesjy (changes requested):

yvonnesjy (approved):

src/js/views/maps/viewfinder/SearchView.js
robyngit: Ah I see, then agree it's not useful!

robyngit:
Without the `viewfinder` styles, the layout breaks:

Screenshot 2024-04-23 at 1 31 00 PM
Screenshot 2024-04-23 at 1 30 53 PM

ianguerin:
> Without the `viewfinder` styles, the layout breaks:
> 
> Screenshot 2024-04-23 at 1 31 00 PM Screenshot 2024-04-23 at 1 30 53 PM

Hmm, yeah couple things wrong there, should be fixed now.

robyngit (approved):

PR #2401, opened 2024-05-03
Add a MapAssets collection to maintain a list of layers in a Map model

ianguerin:
This is just an enhancement to the way we track the layers across layer configurations or layer categories configuration. 
Not urgent.

NCEAS#2400


ianguerin:
Bump on this, review when you can please!

yvonnesjy (approved):

src/js/models/maps/Map.js
```diff
@@ -269,18 +273,22 @@ define([
...
               this.set("layerCategories", assetCategories);
               this.unset("layers");
+              this.set('allLayers', assetCategories.getMapAssetsFlat());
             } else if (isNonEmptyArray(config.layers)) {
-              this.set("layers", new MapAssets(config.layers));
+              const layers = new MapAssets(config.layers);
+              this.set("layers", layers);
               this.get("layers").setMapModel(this);
               this.unset("layerCategories");
+              this.set('allLayers', layers);
```
robyngit, line 282: Why are we setting the same property to both `layers` and `allLayers`?

robyngit (approved):
Looks good! 🎉  Please wait to merge until we have the linting changes in - that will be its own little release.

src/js/models/maps/Map.js
ianguerin: I want a consistent way (regardless of categorized or uncategorized layers) to read a flat list of all of the layers. I could reuse the layers attribute, but right now the code assumes that only one of the two, layers or layerCategories will be populated

robyngit: Thanks for explaining!

robyngit:
@ianguerin feel free to merge now!

PR #2423, opened 2024-05-28
DEMO 2: Github action pass when linting & formatting is ok in changed code

robyngit:
Although the rest of the codebase has linting errors, this PR should pass github action tests because the changed code does not have linting errors.

This is for demo purposes only.


robyngit:
**Note:** The check fails if there are linting errors in any of the *files* changed. Suggestions and PR comments are made only where the linting errors apply to the changed lines of code.

Issue #2373, opened 2024-04-19
Download All doesn't work for specific data package


mbjones:
**Describe the bug**
Clicking "Download All" on this data package results in a new blank tab, but the dataset is not downloaded. The individual file downloads work fine. Issue reported by a researcher in [Redmine issue 27990](https://support.nceas.ucsb.edu/rt/Ticket/Display.html?id=27990)

**To Reproduce**
1. Go to https://search.dataone.org/view/https%3A%2F%2Fpasta.lternet.edu%2Fpackage%2Fmetadata%2Feml%2Fknb-lter-arc%2F10006%2F7
2. Click on 'Download All' or the equivalent icon
3. See new empty tab appear, but no download

**Expected behavior**
The dataset zip file should be present on my disk

**Screenshots**

**Desktop (please complete the following information):**
 - OS: Mac OS
 - Browser: Firefox
 - Version: 124.0.1

**Additional context**

Same problem was seen for other datasets from LTER, ESS-DIVE, and ADC. The third one in this list, if accessed from the ADC site, works fine for download all, and produces the correct zip file.

- https://search.dataone.org/view/https%3A%2F%2Fpasta.lternet.edu%2Fpackage%2Fmetadata%2Feml%2Fknb-lter-nwt%2F264%2F3
- https://search.dataone.org/view/doi%3A10.15485%2F2336865
- https://search.dataone.org/view/doi%3A10.18739%2FA2F18SH0T

It is not clear to me how Download All would work on DataONE, as the download has to redirect to the member repository to get the data. For member repositories like ADC that support the package service, it would be fine to redirect to the ADC package service, but this doesn't work on all nodes. Is it possible that "Download All" should be disabled/not display for DataONE deploys?

This was verified on the newly released MetacatUI 2.28.0, but was originally reported on a prior version in Redmine.

robyngit:
Something change in how we construct these Download All URLs in the recent 2.28.0 release, see the comparison to the previous version:

1. [Package: `https://pasta.lternet.edu/package/metadata/eml/knb-lter-arc/10006/7`](https://search.dataone.org/view/https%3A%2F%2Fpasta.lternet.edu%2Fpackage%2Fmetadata%2Feml%2Fknb-lter-arc%2F10006%2F7)
[...]

PR #2433, opened 2024-06-06
Bugfix 2373 download url

rushirajnenuji:
#2373 

Resolves issues with the package table download URL


robyngit (approved):
Tested with the problematic download buttons listed in the issue, all fixed with this PR! Nice work @rushirajnenuji !! 🎉

Issue #2418, opened 2024-05-17
Nested datasets don't display links and child datasets clearly under the parent dataset


justinkadi:
**Describe the feature you'd like**
For nested datasets, it would be helpful to have the parent dataset include links for all the child datasets nested underneath. In the older UI, there was a hyperlink that could be followed next to the name of each child dataset. Currently, the child datasets are listed as individual folders under the parent dataset.

**Is your feature request related to a problem? Please describe.**
When users are on the parent dataset page, they are unable to find the metadata page for the child datasets. 

**Additional context**
There have been 2 instances where this has come up since the update:

1. [This dataset](https://arcticdata.io/catalog/view/doi:10.18739/A2D795B84) is one of many child datasets nested under this [parent dataset](https://arcticdata.io/catalog/view/doi%3A10.18739%2FA2QJ7801N).
2. This is [another parent dataset](https://arcticdata.io/catalog/view/doi:10.18739/A20Z70X9X) that has this issue. The PI included this request in her email 

> I set this up a few years ago to keep all our datasets under one
header and now some of the other authors of the datasets seemed to
have been left out somehow. It seems there was some kind of change to
the layout of the nested datasets in recent years. Is there a way to
change the author on this tree of datasets? All the datasets are listed under this tree, which was the goal, but
now authorship isn't clearly visible of all the individual datasets.

PR #2434, opened 2024-06-06
Feature 2418 nested package links

rushirajnenuji:
#2418 

Adds a link in the actions panel of the data package table that takes the user to the nested / child dataset landing page


robyngit (changes requested):
Hey @rushirajnenuji! The new link to nested packages looks good, but I'm having trouble getting it working properly on my end.

In the console, I'm seeing the error: `Uncaught ReferenceError: assignment to undeclared variable nestedPackageUrl` for the method `addNestedPackages` in `DataPackageView.js:956`. On line 951, I think there's just a type, and `packageUrl = null;` should be `packageUrl = null,`.

When I change that though, the link for nested packages seems to be leading to the resource map instead of the dataset (e.g. https://arcticdata.io/metacat/d1/mn/v2/views/metacatui/resource_map_doi%3A10.18739%2FA2HX15R89).

I'm seeing this for the package http://localhost:3000/view/doi%3A10.18739%2FA2QJ7801N on the arctic theme.

rushirajnenuji:
Hi @robyngit  - thanks for your review and for catching this error, it seems like I messed up something with the ESLint formatting. This should be fixed with the above two commits and is ready for another review.

robyngit (approved):

robyngit:
Works great now, thanks @rushirajnenuji !

PR #2439, opened 2024-06-07
Do not format helm yaml templates

robyngit:
Prettier misinterprets go templating in yaml as invalid yaml syntax. Unfortunately, there is [not currently support for Helm’s template + yaml language](https://github.com/prettier/prettier/issues/6517), so this PR makes sure Prettier ignore these files. It also updates the helm branch with the latest develop changes, so that Prettier is available to update!


artntek (changes requested):

artntek:
Closing this PR. Instead, @robyngit and I agreed that I shoudl make this change in PR #2427 ([see change here](https://github.com/NCEAS/metacatui/pull/2427/files#diff-b640b344ee7f3f03d2a443795a5d0708ef50e2e6e34214109ab2aad13ad6ba98R10))

PR #2441, opened 2024-06-07
Legend 1: Create Cesium widget container

yvonnesjy:
This container is intended to be a parent for the map and all overlays (ex. legend, scale bar, etc). Reducing the top level container to just the panel and the map will help with the layout migration where we use the top-bottom layout. [Design doc](https://docs.google.com/document/d/1eeS7F_fP53h3tL8gVrDpju-cexvjcDsiwY1gxWsX2EM/edit?usp=sharing)

 If I don't get to migrating the scale bar, I created the tracking issue at: https://github.com/NCEAS/metacatui/issues/2440

### DIFF
--- a/src/css/map-view.css
+++ b/src/css/map-view.css
@@ -433,10 +433,7 @@ represents 1 unit of the given distance measurement. */
 .scale-bar__bar {
   height: 0.25rem;
   transition: width 0.3s ease-in-out;
-  background-color: var(
-    --portal-col-neutral-4,
-    var(--map-col-text__deprecate)
-  );
+  background-color: var(--portal-col-neutral-4, var(--map-col-text__deprecate));
 }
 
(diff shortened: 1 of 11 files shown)


ianguerin:
Is there more functionality coming to the JS side of this in the future? Could this be accomplished with some small modifications to templates/maps/map.html?

yvonnesjy:
> Is there more functionality coming to the JS side of this in the future? Could this be accomplished with some small modifications to templates/maps/map.html?

good qs. this component will indeed be simple with not much actual logic going on. I technically can just update MapView.js, but it feels more correct to me to organize the views a little. Unless there's a strong reason not to, I think one extra container js doesn't hurt.

ianguerin (approved):

robyngit (approved):

PR #2444, opened 2024-06-11
Legend 3: Add legend to map

yvonnesjy:


ianguerin (approved):

src/js/views/maps/MapWidgetContainerView.js
```diff
@@ -34,6 +35,7 @@ define(["backbone", "models/maps/Map", "views/maps/CesiumWidgetView"], (
       /** @inheritdoc */
       render() {
         this.renderMapWidget(this.el, this.model);
+        this.renderLegendContainer(this.el, this.model);
```
robyngit, line 38: Don't need to pass `el` or `model`. Same for mapWidget actually!

src/css/map-view.css
```diff
@@ -1665,3 +1665,99 @@ other class: .ui-slider-range */
...
+.legend-container {
+  z-index: var(--map-z-index-base);
+  position: absolute;
+  left: 1rem;
+  bottom: 1rem;
+  border-radius: var(--map-border-radius-big);
+  background: var(--portal-col-neutral-00, var(--map-col-bkg__deprecate));
+  width: 6.875rem;
+
+  .legend-container__header {
```
robyngit, line 1686: I know metacatui is very inconsistent in our CSS conventions, but we're gradually trying to move toward BEM principles. Would it be a big refactor to move these elements around a bit to make rules less specific? Here, for example, `.legend-container__header` does not need to be nested under `.legend-container` - it gives `legend-container__header` a higher specificity than it needs. With SCSS/SASS, I know we would nest it like so:

```
.legend-container {
  &__header {
    // styles
  }
}
```

which is nicer for readability. But since we're just using plain CSS, I think it would be better to move `.legend-container__header` out of `.legend-container` and just make it a sibling. This way, we can avoid the specificity issue. So:

```
.legend-container {
  // styles
}

.legend-container__header {
  // styles
}
```

What do you think?

src/css/map-view.css
```diff
@@ -1665,3 +1665,99 @@ other class: .ui-slider-range */
...
+
+      .layer-legend__palette {
+        color: var(--portal-col-neutral-8);
+        font-size: 0.8rem;
+        font-weight: 500;
+      }
+    }
+  }
+
+  &.expanded {
```
robyngit, line 1750: I think typically in BEM it would be `legend-container--expanded` but not a big deal

robyngit (changes requested):
This legend is going to be great! 🎉🎉🎉 

A couple minor aesthetic issues I noticed:

On maps without the light theme, e.g. in the catalog, the panels element is weirdly translucent or has a darker text color:
Screenshot 2024-06-17 at 4 26 28 PM

Could we bump the positioning of the legend up on portal maps on smaller screens? See:

Screenshot 2024-06-17 at 4 25 35 PM

src/css/map-view.css
robyngit: ^ If you think it's a good idea, I could make an issue for this and we can address it separately

src/css/map-view.css
yvonnesjy: done!

src/js/views/maps/MapWidgetContainerView.js
yvonnesjy: oh right right!

yvonnesjy:
> in the catalog, the panels element is weirdly translucent or has a darker text color
Fixed

> Could we bump the positioning of the legend up on portal maps on smaller screens?
Done... only if you promise me this code will be cleaned up when layout comes around👀

src/css/map-view.css
yvonnesjy: addressed in #2446 . otherwise the conflicts will be 🫠

robyngit (approved):
🚀🚀🚀

PR #2445, opened 2024-06-11
Legend 4: Render categorical palette

yvonnesjy:


src/js/views/maps/legend/LayerLegendView.js
```diff
@@ -43,27 +44,39 @@ define([
 
       /** @inheritdoc */
       render() {
+        const paletteType = this.model.get("paletteType");
+        if (!paletteType || paletteType === "categorical") {
```
ianguerin, line 48: Can `paletteType` be undefined? Looks like it defaults to "categorical"

ianguerin (approved):

src/js/views/maps/legend/LayerLegendView.js
yvonnesjy: ouioui you're right. removed check

src/js/models/maps/AssetColor.js
```diff
@@ -151,6 +141,12 @@ define(["jquery", "underscore", "backbone"], function ($, _, Backbone) {
             }
           : null;
       },
+
+      /** @returns {string} A string in the format of a css color value. */
```
robyngit, line 145:
```suggestion
      /**
      @returns {string} A string in the format of a css color value.
      @since 0.0.0
      */
```

yvonnesjy:
Ah thanks for catching! Fixed. Is there a way to only mark eslint failure as warning for now?

I also pushed a commit that incorporates the label fixes
- `<layer name> (<property label || property name>)`: Not reflected in mock yet but I got a 👍🏻 from Doug on this formatting
- `<color label> || <color value>` for categorical swatch

PR #2446, opened 2024-06-11
Legend 5: Render continuous palette

yvonnesjy:


src/js/views/maps/LegendView.js
```diff
@@ -361,218 +331,182 @@ define([
...
-                    .tooltip("show");
+      renderContinuousPreviewLegend(colorPalette) {
+        if (!colorPalette) {
+          return;
+        }
+        const view = this;
+        // Data to use in d3
+        let data = colorPalette.get("colors").toJSON();
+        // The max width of the SVG
+        const { width } = this.previewSvgDimensions;
```
ianguerin, line 342: Can combine with assignment below

src/js/views/maps/LegendView.js
```diff
@@ -361,218 +331,182 @@ define([
...
+              const relativePosition = xMouse / width;
+              // Get the value at the relative position by interpolating the data
+              let value = d3.interpolate(
+                data[0].value,
+                data[data.length - 1].value,
+              )(relativePosition);
+              // Show tooltip with the value
+              if (value || value === 0) {
+                // Round or show in scientific notation
+                if (roundingConstant) {
```
ianguerin, line 432: This duplicates logic from the helper, can we use the Utilities function instead?

src/js/common/Utilities.js
```diff
@@ -102,21 +94,56 @@ define(["jquery", "underscore"], function ($, _) {
...
-      names = names.filter(function (name) {
-        return name.length > 0;
-      });
+      names = names.filter((name) => name.length > 0);
 
       return names;
     },
+
+    // Format the number into a string with better readability.
+    formatNumber(value, min, max) {
```
ianguerin, line 110: I wrote an expectation that failed:
```
expect(EntityUtils.formatNumber(0.99999, 0, 0.0001)).to.equal("0.0001");
```
I would expect min and max to define the range, but that doesn't appear to be the case, what does min and max represent?

ianguerin (approved):

src/js/common/Utilities.js
```diff
@@ -102,21 +94,56 @@ define(["jquery", "underscore"], function ($, _) {
...
+      if (!value && value !== 0) {
```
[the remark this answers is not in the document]
yvonnesjy: value === null, undefined, "", and all other silly cases. changed to typeof() === "number" since that's closer to what's actually required here.

src/js/common/Utilities.js
```diff
@@ -102,21 +94,56 @@ define(["jquery", "underscore"], function ($, _) {
...
+        return "";
+      }
+
+      const roundingConstant = Utilities.getRoundingConstant(max - min);
+      if (roundingConstant) {
+        return (
```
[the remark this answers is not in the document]
yvonnesjy: yup! updated

src/js/views/maps/LegendView.js
yvonnesjy: I'm putting minimum effort into LegendView because it's being removed in #2448 . Just pushed a commit that reverts all manual lint fixes I did here since I ran into some issues with d3.

src/js/views/maps/LegendView.js
yvonnesjy: Same as above- the goal for LegendView here is to maintain a noop with minimum effort

src/js/common/Utilities.js
yvonnesjy: Ah yea, the min and max here is misleading. In practice, value should always be in between min and max. And the method only cares about (max - min) to decide how many decimal places to use. I updated the param to just use a `range` instead, and added more details to the JSDocs.

src/js/views/maps/LegendView.js
ianguerin: Ack

src/js/common/Utilities.js
```diff
@@ -102,21 +94,66 @@ define(["jquery", "underscore"], function ($, _) {
...
 
       return names;
     },
+
+    /**
+     * Format the number into a string with better readability, based on the manitude of a
+     * range this number falls in.
+     * @param {number} value The number value to be formatted.
+     * @param {number} range The range of numerics this value can fall in.
+     * @returns {string} A formatted number based on the magnitude of `range`.
```
robyngit, line 114:
```suggestion
     * @returns {string} A formatted number based on the magnitude of `range`.
     * @since 0.0.0
```

src/js/common/Utilities.js
```diff
@@ -102,21 +94,66 @@ define(["jquery", "underscore"], function ($, _) {
...
+      if (numDecimalPlaces !== null) {
+        return value.toFixed(numDecimalPlaces);
+      }
+      return value.toExponential(2).toString();
+    },
+
+    /**
+     * Calculate the number of decimal places we should use based on the range of the data.
+     * @param {number} range The range of data values.
+     * @returns {number} The number of decimal places we should use.
```
robyngit, line 134:
```suggestion
     * @returns {number} The number of decimal places we should use.
     * @since 0.0.0
```

src/js/common/Utilities.js
```diff
@@ -102,21 +94,66 @@ define(["jquery", "underscore"], function ($, _) {
...
+        return 3; // Allow 3 decimal places
+      }
+      if (range < 1) {
+        return 2; // Allow 2 decimal places
+      }
+      if (range <= 100) {
+        return 1; // Allow 1 decimal places
+      }
+      return 0; // No decimal places
+    },
```
robyngit, line 156: 👍🏻 Good idea moving this to Utilities

robyngit (approved):

PR #2448, opened 2024-06-13
Legend: Remove legend previews from toolbar

yvonnesjy:
Deleting the previews as the [UX mock](https://www.figma.com/design/6bbLaj8VZViuArkhpzMm9M/PDG-Enhancements?node-id=42-2041&t=moYEH5sCJgehbqL6-4) suggests. [This slack thread](https://pdgateway.slack.com/archives/C01BNVBLFML/p1718148582249159) has also acked the change.


src/js/views/maps/LayerDetailsView.js
```diff
@@ -143,197 +134,153 @@ define([
...
-            }
-          }
-        } catch (e) {
-          console.log(
-            "A LayerDetailsView failed to initialize. Error message: " + e,
-          );
+      initialize(options) {
+        // Get all the options and apply them to this view
+        if (typeof options === "object") {
+          Object.entries(options).forEach(([key, value]) => {
```
ianguerin, line 159: I think `Object.assign(this, options)` works in the same way; shorter, just as readable?

ianguerin (approved):

src/js/views/maps/LayerDetailsView.js
yvonnesjy: like! done

robyngit:
@yvonnesjy do we also want to remove the little image preview from the base layers and other imagery layers? Has there been consideration about keeping those or moving them somewhere else (legend? in place of icon?)

yvonnesjy:
> @yvonnesjy do we also want to remove the little image preview from the base layers and other imagery layers? Has there been consideration about keeping those or moving them somewhere else (legend? in place of icon?)

Hmmm good questions. @hngrtr ?

Issue #2454, opened 2024-06-18
Change appConfigPath in index.html to the relative path config/config.js


robyngit:
To accommodate the planned structure in [Kubernetes deployments](https://github.com/NCEAS/metacatui/issues/2368), we need to change the `appConfigPath` in `index.html` from an absolute path to a relative one.

The [current path](https://github.com/NCEAS/metacatui/blob/795d778d72e9e386c9cb0430dd38efdbf16f15ef/src/index.html#L7C5-L7C45), `var appConfigPath = "/config/config.js"`, is a problem for metacatui deployments that are in a sub-directory of a domain, like https://arcticdata.io/catalog. Currently, this deployment updates the config path in index to prepend `/catalog/` to the start of the path, but this would not be needed if we use the relative path instead.

We should ensure that this change will be compatible with all existing (non Kubernetes) installations.

See also #2453

rushirajnenuji:
Hi @robyngit  and @artntek - 

Ran into the issue with the [local deployment of metacatUI](https://nceas.github.io/metacatui/install/local.html), with the most recent develop branch of MetacatUI. 

It looks like `MetacatUI` runs into issues with getting the config when you directly navigating links in the app.

Error:
```
The loader.js file failed to load. The MetacatUI `root` setting is probably incorrect. Set a `root` path for MetacatUI in the AppConfig.
```

To reproduce,
- checkout develop branch
- set up config:
```
MetacatUI.AppConfig = {
    root: "/",
    theme: "arctic",
}
```
- run local npm server
- visit any page `https://localhost:3000/view/{PID}` or `https://localhost:3000/portals/{portal_label}`

robyngit:
Thanks for finding this @rushirajnenuji ! You're right, this solution isn't going to work. I'm going to re-open this issue and revert the change in develop... @artntek we'll need to figure out another way to do this

PR #2463, opened 2024-06-19
change /config/config.js to config/config.js

artntek:
see #2454 - Change appConfigPath in index.html to the relative path config/config.js

#### NOTE: BEFORE MERGING
...we should ensure that this change will be compatible with all existing (non Kubernetes) installations.


robyngit:
- [x] Updated our current upgrade script to match and replace `/config/config.js` OR `config/config.js` (https://github.nceas.ucsb.edu/NCEAS/metacatui-config/issues/1)

artntek:
For testing, I created k8s deployments for knb, arctic, and cerp, all using the relative path in this `feature-2454-relative-appConfigPath` branch. What additional verification do you recommend before merging, @robyngit?

robyngit (approved):

PR #2467, opened 2024-06-24
Legend: Bug bash fixes

yvonnesjy:
1. Take into consideration of "min"/"max" as the value of a continuous palette. The exact values of these fields can change based on the viewport.
2. Lower the max height of legend now that we've added the share link button to toolbar link.


ianguerin (approved):

robyngit:
Fantastic @yvonnesjy !!! 🎉

robyngit (approved):

PR #2488, opened 2024-07-23
Added clickFeatureAction config to the Layer Level

alonakos:


src/js/models/maps/assets/MapAsset.js
```diff
@@ -108,6 +108,9 @@ define([
        * from the layer list.
        * @property {boolean} [showOpacitySlider = true] Set to true to show opacity slider
        * for the layer.
+       * @property {boolean} [clickFeatureAction = null] The action to take when a user clicks on a feature on the layer. The
```
robyngit, line 111:
```suggestion
       * @property {"showDetails"|"zoom"} [clickFeatureAction = null] The action to take when a user clicks on a feature on the layer. The
```

src/js/views/maps/LayerDetailView.js
```diff
@@ -58,6 +58,10 @@ define([
        * @property {string} label The element that holds the view's label text
        * @property {string} contentContainer The container into which the contentView's
        * rendered content will be placed
+       * @property {string} [clickFeatureAction] - The default
+       * action to take when a user clicks on a feature on the layer. The
+       * available options are "showDetails" (show the feature details in the
+       * sidebar) or "zoom" (zoom to the feature's location).
```
robyngit, line 64: I think this was committed by accident?

robyngit (changes requested):
**Great work on the click feature implementation!** I see you've effectively used optional chaining (`.?`) and handled different scenarios well. I tested setting the `clickFeatureAction` to `"showDetails"` on the Geohash layer in the [config for the map in the arctic theme](https://github.com/NCEAS/metacatui/blob/bcc2c7e83b64e72789aa392de4beeacd206adf8a/src/js/themes/arctic/config.js#L105-L152) locally. I just temporarily added a line around [here](https://github.com/NCEAS/metacatui/blob/bcc2c7e83b64e72789aa392de4beeacd206adf8a/src/js/themes/arctic/config.js#L118) with `clickFeatureAction: "showDetails",` but it didn't seem to work. When I clicked a geohash tile, the map did not zoom, but it did not show the details panel either.

I believe it's becuase we are also checking for the value of the Map's `clickFeatureAction` in the `MapView` [here](https://github.com/NCEAS/metacatui/blob/bcc2c7e83b64e72789aa392de4beeacd206adf8a/src/js/views/maps/MapView.js#L133-L134). We only render the feature details panel if the `clickFeatureAction` is set to `"showDetails"` at the map level! The details panel is hidden until a feature is clicked on the map, then the contents of the panel is updated and the panel is shown. Before, we were expecting that this would never happen if the map is set to `"zoom"`, but now we can have the details panel show up if any layer in the map is set to `"showDetails"`. 

With these changes, we can render the panel either way and not check anymore what the map-level `clickFeatureAction` is set to. Could you please remove the check for the map-level `clickFeatureAction` in the `MapView` and test if the details panel shows up when the Geohash layer is set to `"showDetails"`? I'd be happy to walk through this idea and discuss it further if you'd like!

src/js/models/maps/MapInteraction.js
```diff
@@ -28,7 +28,7 @@ define([
    * @since 2.27.0
    * @extends Backbone.Model
    */
-  var MapInteraction = Backbone.Model.extend(
+  let MapInteraction = Backbone.Model.extend(
```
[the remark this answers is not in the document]
alonakos: Thank you for noticing!

src/js/views/maps/LayerDetailView.js
alonakos: removed, thank you!

robyngit (approved):
Wonderful!! 🎉 Reviewed & tested and it all looks great. Thanks for working on this @alonakos!!

Issue #2487, opened 2024-07-18
Change initContainer code to manipulate path for config.js


artntek:
Our attempt to use a relative path for `config.js` was unsuccessful - see #2454 

The only viable alternative at this stage is to edit this line in `index.html` when installign metacatui:

```javascript
    var appConfigPath = "/config/config.js";
```

Since this file is also used for non-k8s deployments, we can't add go templating.

We already have an `initContainer` defined in `deployment.yaml`, so it should be pretty straightforward to add code there in order to inser the root path before the beginning of the `appConfigPath` shown above.

artntek:
resolved by PR #2489

PR #2489, opened 2024-07-25
Change initContainer code to manipulate path for config.js

artntek:
see #2487


rushirajnenuji (approved):
Thanks @artntek  - for adding these fixes in.

PR #2499, opened 2024-08-16
Bugfix-2498-added checks for model + fixed syntax in the old code

alonakos:
This PR resolves the issue introduced in the previous commit. It also addressed ReviewDog and lining problems that were present in the old code.

### DIFF
--- a/src/js/models/maps/Map.js
+++ b/src/js/models/maps/Map.js
@@ -273,6 +273,19 @@ define([
        * used instead (see {@link MapModel#defaults}).
        */
       initialize(config) {
+        config.layers = [
+          {
+            "label": "Local Stories",
+            "layerId": "ls",
+            "type": "GeoJsonDataSource",
+            "opacity": 0.75,
+            "showOpacitySlider": false,
+            "clickFeatureAction": "showDetails",
+            "cesiumOptions": {
+              "data": "https://www.leonetwork.org/en/explore/posts?query=&type=TWEET&type=POST&type=ARTICLE&mode=geojson_compact&region=&polygon=&bbox=&minlat=&maxlat=&near=&radius=&categories=PERMAFROST%7cPermafrost+Change&categories_anyOrAll=ANY&fromdate=&todate="
+            }
+          }
+        ]
         try {
           if (config && config instanceof Object) {
             if (isNonEmptyArray(config.layerCategories)) {
@@ -395,7 +408,8 @@ define([
       getLayerGroups() {
         if (this.has("layerCategories")) {
           return this.get("layerCategories").getMapAssets();
-        } else if (this.has("layers")) {
+        }
+        if (this.has("layers")) {
           return [this.get("layers")];
         }
         return [];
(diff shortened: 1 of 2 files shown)


src/js/views/maps/LayerDetailsView.js
```diff
@@ -295,9 +292,8 @@ define([
           return this;
         } catch (error) {
           console.log(
```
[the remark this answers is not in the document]
alonakos: I have seen these console.log statements after catch all over the code, so I did not remove any in this file. @robyngit do we want to keep them?

src/js/models/maps/Map.js
```diff
@@ -395,7 +395,8 @@ define([
       getLayerGroups() {
         if (this.has("layerCategories")) {
           return this.get("layerCategories").getMapAssets();
-        } else if (this.has("layers")) {
+        }
```
alonakos, line 398: no need for `else`" after `return`

src/js/views/maps/LayerDetailsView.js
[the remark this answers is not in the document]
robyngit: @alonakos yes, good question!!

Most of these were my doing back in the day!! We used to have an issue where tiny mistakes in one part of the codebase would crash the entire app. So we decided to use more `try-catch`es to catch these errors, but I was over zealous. While we should certainly still catch errors to handle them gracefully, it's not needed all over the place like it is now. Also, we should do something more useful than `console.log`ing the error statement.

Also, since we're moving toward the AirBnB style guide, we want to avoiding using `console.log` in production code. So I would say remove the `try-catch`es and `console.log`s when they are redundant or at least not very useful. When it makes sense, we should keep the `try-catch` and add a useful fallback behaviour in the `catch` part.

src/js/models/maps/Map.js
```diff
@@ -395,7 +395,8 @@ define([
...
+        if (this.has("layers")) {
           return [this.get("layers")];
```
robyngit, line 400: @alonakos curious why you changed this from `else if` to `if`? Was this part of the bug fix?

src/js/views/maps/LayerDetailsView.js
```diff
@@ -210,21 +208,21 @@ define([
...
           );
-          const labelEl = this.el.querySelector("." + this.classes.label);
+          const labelEl = this.el.querySelector(`.${this.classes.label}`);
 
           // Render each section in the Details panel
           this.renderedSections = _.clone(this.sections);
 
           // Remove and do not render opacity section if showOpacitySlider is false
-          if (model.get("showOpacitySlider") === false) {
+          if (model && model.get("showOpacitySlider") === false) {
```
robyngit, line 219: this is great, just FYI a shortcut for this is ` model?.get("showOpacitySlider")`

alonakos: Thank you @robyngit! I changed it

src/js/views/maps/LayerDetailsView.js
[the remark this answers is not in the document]
alonakos: Sounds good! I opened a new issue here: https://github.com/NCEAS/metacatui/issues/2500

robyngit (approved):
🎉 Tested it out, no more errors, thank you! :)

PR #2507, opened 2024-08-29
Show appropriate message for "processing" status on Metadata Quality page

robyngit:
Previously we were displaying the "in queue" message only when the status was set to "QUEUED". The "PROCESSING" status resulted in a generic error message that was not very informative. Possible statuses that the view looks for now include (all case insensitive):

- SUCCESS
- FAILURE
- QUEUED
- PROCESSING
- 404 (from `fetchResponse.status`)

@jeanetteclark are there any other possible statuses we should be aware of?

Note: This PR depends on changes from https://github.com/NCEAS/metacatui/pull/2506 and should not be merged until that PR has been merged and this PR has been rebased with develop.

### DIFF
--- a/src/js/views/MdqRunView.js
+++ b/src/js/views/MdqRunView.js
@@ -28,7 +28,7 @@ define([
   const MSG_ERROR_GENERATING_REPORT =
     "There was an error generating the assessment report.";
   const MSG_QUEUED_REPORT =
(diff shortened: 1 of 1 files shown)


rushirajnenuji (approved):
Reviewed both #2506 and #2507 and they both look great! Thank you for working on this @robyngit and adding this support to MetacatUI. Looking forward to Metadig getting the markdown response in its next release.

Issue #2395, opened 2024-05-01
Issues with Citation Display


rushirajnenuji:
**Describe the bug**
The summary pages in MetacatUI errors out while trying to render citations

**To Reproduce**
1. Go to 'https://arcticdata.io/catalog/profile' (or any metrics page)
2. Scroll tp the Citations section on the stats / metrics page

**Expected behavior**
Render complete list of Citaitons.

**Screenshots**
<img width="1439" alt="Screenshot 2024-05-01 at 10 54 34" src="https://github.com/NCEAS/metacatui/assets/19696935/bc85edae-4e47-4f72-8885-1c4dbdd4b820">

**Desktop (please complete the following information):**
 - OS: macOS 14
 - Browser Chrome (Version 124.0.X) and Firefox (Version 125.0.3)

**Console logs**
<img width="1288" alt="Screenshot 2024-04-29 at 10 52 38" src="https://github.com/NCEAS/metacatui/assets/19696935/db39435b-f4a4-42ca-8320-cce6c6353fd3">

mbjones:
@rushirajnenuji @robyngit I'm still seeing this problem on FF 126.0.1 on MacOS, for ADC and ESS-DIVE at least.

<img width="1912" alt="image" src="https://github.com/NCEAS/metacatui/assets/766407/8420861d-8ba6-4987-bd35-1d88adfbf5db">

rushirajnenuji:
Update: 

- Jim ran into this error when looking at ESS DIVE Citations. 
![image](https://github.com/user-attachments/assets/60da8d92-f95d-4bd3-bc5b-a5bbec4215d5)

Jim saw these issues on the following browser versions:
```
Chrome 127.0.6533.100 on my MBP
Safari 17.5 on my MBP
Current Safari on my iPhone (iOS 17.5.1)
```

- jeanetteclark was able to see the citations displayed with the older FF version
![Screen Shot 2024-08-26 at 11 58 43 AM](https://github.com/user-attachments/assets/6cb22506-0d56-4619-8504-d6e143ebb7f7)
```
Firefox 128.0.3
```

I verified that the [fix above](https://github.com/NCEAS/metacatui/commit/e2e9b7c862200ff3993ddf5f0372641f7d8b3463) resolves the issue for my local development instance (on both latest Chrome and FF).

Next steps:
Test some more and PR. Hoping to get this in for the `2.31.0` release

PR #2510, opened 2024-09-05
Bugfix 2395 citations display

rushirajnenuji:
Fixes issues with Citation display on dataset and repository landing pages.
#2395


robyngit (approved):
🪄 you fixed what I broke! Sorry about that unnecessary circular dependency and thanks for figuring this out!!!

PR #2512, opened 2024-09-09
Show coords that are zero in metadataIndexView

### LINKED ISSUE #2505: Bounding coordinates set to 0 don't display on dataset landing pages rendered from the index
The bounding coordinates are not displayed correctly on dataset landing pages that are rendered from the index when coordinates are set to 0. See for example, [this dataset](https://search.dataone.org/view/10.24431/rw1k329) which has a southBound coord set to 0. The south coordinate is not shown and the Google Map preview is broken.

I assume the problem occurs because the code incorrectly treats 0 as a falsy value, leading to the coordinates not being shown. We are doing that [here](https://github.com/NCEAS/metacatui/blob/db1a1e0e9c8dad9c2980f940de4620503e09771b/src/js/views/MetadataView.js#L3266) for example, and repeating the same check elsewhere.

Seems like we could use a function like:
```js
const allCoordsExist = () => {
  const coordAttrs = ["northBoundCoord", "eastBoundCoord", "southBoundCoord", "westBoundCoord"];
  return coordAttrs.every(attr => model.get(attr) !== undefined);
}
```

robyngit:
Fixes #2505

See the difference in geographic coverage for this dataset: https://search.dataone.org/view/10.24431/rw1k329


rushirajnenuji (approved):
✅

PR #2532, opened 2024-09-20
Allow iFrames from trusted sources in MarkdownView

robyngit:
- Update [Showdown.js](https://showdownjs.com/) (markdown renderer) to the latest version
- Allow iFrames through the xss filter with limited attributes so that they can be processed by the new Showdown extension
- Add a new Showdown extension to allow iFrames from trusted urls, and filter out those from any other url
- Add a property in the AppModel to store the list of trusted sources. These urls can contain wild cards, e.g. `https://*dataone.org*`
- Also add a sandbox attribute to the iFrame to limit the capabilities of the iFrame
- Add the new Showdown extension to the MarkdownView

### 🧪 Testing

Currently, iframes can be added to markdown by adding the HTML code directly, e.g.:
```markdown
<iframe src="www.youtube.com/embed/xvFZjo5PgG0?si=wLUW0M4qF4brYbeS" width="560" height="315"></iframe>
```

This portal has a bunch of iframes and other markdown for testing purposes: https://demo.arcticdata.io/portals/markdown-tests


rushirajnenuji (approved):
Tested adding iFrames to portals from sources in the allow list, and the showdown extension works as expected. Thank you for getting this feature in @robyngit  and my apologies for getting late to this PRs.

Issue #2538, opened 2024-09-23
Allow dropdowns for `sem_annotation` field in the Filter Editor


robyngit:
- The Filter Editor restricts filters that include the `sem_annotation` field to only the "free text" UI filter, (details in issue #1900).
- We should allow dropdowns for the `sem_annotation` field (like so: c58ffc78d9d24431f1b8027744b0314753b11600).
- Users should be able to select terms for the dropdown using the BioOntology widget (more complex, requires additional work).
- This issue arose when @justinkadi was trying to add a new filter to the [Arctic Report Card Data Portal](https://arcticdata.io/catalog/portals/reportcard/Data) for 2023 Essay topics (e.g., Precipitation, Subsea Permafrost)

PR #2539, opened 2024-09-24
Allow dropdown filter UI for `sem_annotation` field in portal editor

robyngit:
Fixes issue #2538


robyngit:
Self-reviewing because this is _very_ minor code change

PR #2546, opened 2024-10-03
Show canonical dataset in MetadataView

robyngit:
This PR introduces a feature that a visually identifies a dataset as essentially a duplicate of another in the `MetadataView`. It does this by adding and modifying some components of the view.

### New `CanonicalDatasetHandlerView`

Added a new view, `CanonicalDatasetHandlerView` that handles identifying and displaying the canonical dataset within a Metadata view.

This new view is a "scoped subview", a self-contained view that is responsible for managing a specific part of UI within a larger parent view. In this case, the `CanonicalDatasetHandlerView` handles inspecting and modifying specific parts of the already-rendered DOM within the `MetadataView`. Creating a new view for this purpose avoids adding more complexity to the already very large `MetadataView`.

The `CanonicalDatasetHandlerView` is responsible for the following tasks:
* Inspects the annotations in the `MetadataView` to identify a canonical dataset.
  * When annotations include both `schema.org:sameAs` and `prov:derivedFrom` annotations that have the value, this is assumed to be a reference the DOI of the canonical dataset.
* Modifies the `MetadataView` when a canonical dataset is detected, by:
  * Hiding the annotation views that identify the canonical dataset
  * Adding a "info icon" that indicates the dataset is essential a duplicate of another dataset. (An info icon is like the "private" or "archived" icons near the title).
   * Adds a second citation to the citation modal
   * Adds a blue alert box at the top of the page, under the title, that includes the citation to the original dataset

canonical screenshot

### New `CrossRef` model
* Introduced a new model to handle querying the CrossRef API for metadata about a DOI, including methods for caching responses and clearing the cache.

#### Citation Model
* Added properties for `issue` and methods to extract `volume` and `issue` from the source model. Updated `getJournalFromSourceModel` and `getPidFromSourceModel` to handle getting these values from a `CrossRefModel`.

#### Citation View
* Updated the template to include `issue` and adjust the formatting of the journal, volume, and page information.

#### Metadata View
* Deprecated the metadataInfoIcons template in favour of rendering icons directly in the MetadataView methods. This allows us to reuse the logic for adding a "duplicate" icon in the MetadataView along side the private and archived icons.
* Added checks to avoid re-rendering issue
* Insert and render the `CanonicalDatasetHandlerView`


robyngit:
Also fix a problem found in testing:

[ ] datasets that are private but that the logged in user has authorization to view are sometimes shown as private anyways.

rushirajnenuji (approved):
Hi @robyngit, this looks great! 🎉

It is great to have the functionality to query Crossref! Tested this on demo ADC and the new elements show up for datasets with source information.  

I ran into some unit tests failing, but @robyngit has already added a fix for the same here - https://github.com/NCEAS/metacatui/issues/2557

PR #2551, opened 2024-10-15
Allow adding, removing, and changing the "canonical dataset" in the Editor

robyngit:
Add an input in the Overview page of the EML Editor View that allows users to identify the authoritative version of the dataset:

image


robyngit:
Left to do:
- [ ] Validate the URI value and show validation errors
- [ ] Update the `metadataOverview` template that is replicated in other themes

rushirajnenuji (approved):
Reviewed and tested this feature on my local instance of metacatUI. I was able to add, view and modify source dataset annotations, these changes looks good to me. Thank you @robyngit

PR #2556, opened 2024-10-21
Add email validation to EML Party model

### LINKED ISSUE #2554: Email validation does not prevent dotless domains
**Describe the bug**
Metadata input form does not prevent dotless domains (e.g., "user@com") from being accepted as valid email addresses. While dotless domains are possible for emails, they are prohibited by ICANN and can cause email delivery issues with SMTP.

**To Reproduce**

1. Go to Metadata Landing page Editor
2. Input contact email with dotless domain characters (e.g., "fakeuser@test")
3. Fill in other required fields
4. Submit form

**Expected behavior**
The system should reject the email address with a dotless domain and display an error message indicating that the domain is not valid.

**Additional context**
Dotless emails are not accepted by OSTI ELINK2.0 and will create issues for the publication and reservation of DOIs.

robyngit:
- Include unit tests for email validation in spec file
- Also add extra pattern matching in the EML Party template

Fixes #2554

### DIFF
--- a/src/js/models/metadata/eml211/EMLParty.js
+++ b/src/js/models/metadata/eml211/EMLParty.js
@@ -886,7 +886,7 @@ define(["jquery", "underscore", "backbone", "models/DataONEObject"], function (
           this.get("type") == "associatedParty"
             ? this.get("roles")
             : [this.get("type")];
-        for (role of roles) {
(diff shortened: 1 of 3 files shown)


robyngit:
These are minor changes that are okay without a major code review.

PR #2559, opened 2024-10-23
Move JSON-LD schema.org logic into single view + model; truncate or pad descriptions that are too long or too short

### LINKED ISSUE #1899: Truncate description field in injected JSON-LD to 5000 characters
@mbjones noted a few of our datasets on arcticdata.io were failing Google's `schema:Dataset` validation due to "Invalid string length in field "description". When we go into the Search Console, we can see that Google _is_ seeing the `description` field. We then noticed that Google specifies that the `description` field should be between 50 and 5000 characters:

> The summary must be between 50 and 5000 characters long.

From: https://developers.google.com/search/docs/advanced/structured-data/dataset#dataset

In at least one of the offending documents, the abstract is 5779 characters so the above rule seems like a likely cause for the failed validation. Let's truncate the description field to 5000 characters for all records and hopefully these errors will go away.

robyngit:
This PR introduces a new `SchemaOrgView` and `SchemaOrg` mode, updates views to use these new components, and removes redundant JSON-LD generation code. With management of JSON-LD centralized, added a check for the length of the description in the Dataset json-ld to adhere to Google's guidelines. It either truncates the description if it's above 5000 characters, or adds some text if the description is less than 50 characters.

In working on this, I found a few issues with the current JSON-LD generation that I also fixed in this PR:

- The app config option `isJSONLDEnabled`, when false, now prevents *all* instances of schema.org
  json-ld from being added in the app; previously it only applied to the Dataset json-ld, which seems to have been an oversight.
- Fix issue where DataCatalog json-ld was empty if node metadata hadn't synced yet
- Add missing DataCatalog json-ld for the DataCatalogView. Previously there was a method to do this, but it was never called.
- Use the theme's jsonLD template as the default json-ld content when the current page doesn't have an specific schema.org json-ld content (i.e. it's not a Dataset landing page and not a DataCatalog). Previously, this content would only be shown on the very first page that a user lands on (and only if it wasn't a DataCatalog page or a Dataset landing page), then would be removed upon navigating to another page.
- Fix a syntax error in the KNB theme's json-ld template. There was an extra `}` at the end of the template that was causing a parsing error.

Fixes #1899


robyngit:
The branch is currently deployed at https://metacatui.test.dataone.org/, pointing to the ⚠️ **production** ⚠️ ADC node

maggieklope (approved):
Datasets on metacatui.test.dataone.org with both too long and too short abstracts meet the description requirement of between 50 and 5000 characters. Thanks Robyn!

rushirajnenuji (approved):
LGTM! Thank you, @robyngit. I remember that this was one of my first feature implementations while getting familiar with MetacatUI, and Bryce and Lauren were helping me figure out how to get this in. It is great to see all the changes and improvements that have come along over the years. Thank you again, Robyn! 🙏🏻🎉

Issue #2562, opened 2024-11-04
Helm chart congfig.js template interprets JS Object declarations as Strings


artntek:
found when adding banner to test.adc site. These lines in values.yaml:
```yaml
  temporaryMessageEndTime: new Date("2222-06-16T13:30:00"),
  temporaryMessageStartTime: new Date("2020-06-16T13:30:00")
```
get quoted in `config.js:`
```javascript
  temporaryMessageEndTime: "new Date(\"2222-06-16T13:30:00\")",
  temporaryMessageStartTime: "new Date(\"2020-06-16T13:30:00\")",
```
...when in fact they should be written like this:
```javascript
  temporaryMessageEndTime: new Date("2222-06-16T13:30:00"),
  temporaryMessageStartTime: new Date("2020-06-16T13:30:00"),
```

artntek:
reopening to Tag these changes as metacatui helm chart-1.0.0 release

PR #2563, opened 2024-11-04
Bugfix 2562 helm js obj quoting

artntek:
don't quote JS object declarations in config.js

see #2562


robyngit (approved):
Looks like a good regex to fix the `new Date` problem :+1:. We might want to look through the config to figure out which other options could be a problem, other than those that might use object instantiation. Or maybe we could not add quotes and depend on the yaml config to add quotes to strings and leave them out for JS code? Just some ideas. Will merge, thanks for fixing!!

PR #2576, opened 2024-12-04
Validate orcids entered in the Editor

### LINKED ISSUE #2553: Incorrectly formatted ORCID are accepted
**Describe the bug**
We've encountered an issue where users can submit incorrectly formatted ORCID. Specifically, the system currently accepts ORCIDs with more than 16 digits (e.g.  “https://orcid.org/0000-0000-0001-9244-1708” rather than “https://orcid.org/0000-0001-9244-1708”) 

**To Reproduce**

1. Go to Metadata Landing page Editor
2. Input contact ORCID with additional characters (e.g., "https://orcid.org/0000-0000-0001-9244-1708")
3. Fill in other required fields 
4. Submit form

**Expected behavior**
The system should reject the ORCID with additional characters and display an error message indicating the correct format.

**Additional context**
Malformed ORCIDs are not accepted by OSTI and cause issues for the publication and reservation of DOIs

robyngit:
This PR adds a new orcid validation method as well as unit tests for that method.

This prevents users from saving orcids in the "People" section of the editor that are not valid.

Valid ORCIDs include:
- Full URL (http): http://orcid.org/0000-0000-0000-0000
- Full URL (https): https://orcid.org/0000-0000-0000-0000
- Numbers only, with hyphens: 0000-0000-0000-0000.

source: [orcid.org](https://support.orcid.org/hc/en-us/articles/17697515256855-I-entered-my-ORCID-iD-in-a-form-and-it-said-it-s-invalid)

However, when the EML is serialized, the ORCID will be standardized to the https format.

Note that the field for entering ORCIDs in the editor is actually the "userID" field in EML, and this technically allows for any sort of user ID. So ID strings which don't include "orcid.org" and which do not exactly follow the `0000-0000-0000-0000` format are assumed to be another type of user ID and are assumed to be valid.

To test these changes, try entering different orcids, both valid and invalid in the Editor. Also try saving. A user should not be able to save a dataset with an invalid orcid.

image

Closes #2553


rushirajnenuji (approved):
Thanks @robyngit , this looks good

PR #2586, opened 2024-12-13
bug: Fixes handing large number data files in Metadata Editor

vchendrix:
#### Summary
This pull request addresses two issue tickets related to handling large numbers of files when editing a dataset. One issue involves users uploading tens to hundreds of files at once, which inundates Metacat and leads to unrecoverable upload failures, rendering the dataset unsavable. The other issue pertains to loading the metadata editor when a dataset contains a large number of files. The former fix batches the upload and provides a configuration for `batchSizeUpload`. The latter batches the loading of the system metadata and prevents the user from saving until all data file metadata is loaded, with the batch size configurable via `batchSizeFetch`. Additionally, this pull request implements batch file upload functionality using promises and fixes issues related to loading a large number of data files in the Metadata Editor.

_This fix is a high priority for ESS-DIVE and Wildland Fire Science Intitiative as it significantly improves user experience and it will prevent data corruption while handling a large number of files in datasets_

#### Features and Fixes

**Batch File Upload with Promises**
- **Added `batchSizeUpload` Configuration in `AppModel.js`**: Introduced a new configuration to control the batch size for file uploads.
- **Implemented `uploadFilesInBatch` Method in `DataItemView.js`**: Added a method to handle batch file uploads using promises, ensuring that the `_.each` loop completes before proceeding to batch processing.

**Fixes for Loading Large Number of Data Files**
- **Added `fetchMemberModels` Method to `DataPackage`**: Introduced a method to fetch member models in batches.
- **Updated `fetch` Method in `DataPackage`**: Modified the `fetch` method to use the new `fetchMemberModels` method.
- **Added Listener for `numLoadingFileMetadata` Change in `EML211EditorView`**: Added a listener to handle changes in the number of loading file metadata.
- **Updated `toggleEnableControls` in `EML211EditorView`**: Enhanced the method to handle `numLoadingFileMetadata`.
- **Added `fetchBatchSize` Configuration to `AppModel`**: Introduced a configuration to control the batch size for fetching member models.

#### Testing
This fix has been tested on MacOS using the latest versions of Chrome, Firefox, and Safari.

#### Issues Closed
- Closes nceas/metacatui#2224
- Closes nceas/metacatui#2547

PR #2596, opened 2025-01-28
Fix wording on editor button when metadata loading

robyngit:
Before: waiting for 100 file metadata to load...
After: waiting for metadata from 100 files to load....

Issue #2586


robyngit:
Small change, skipped review

PR #2608, opened 2025-02-05
fix: improves DataPackage fetch method

vchendrix:
- Refactored the fetch method in DataPackage to return a Promise and handle options correctly.
- Added detailed documentation for fetch method usage.
- Improved error handling and retry logic for fetch method.
- Added unit tests for DataPackage fetch method to ensure proper functionality.
- Fixes the issue of a new resource map ID not being created on update

Closes #2607

PR #2613, opened 2025-02-14
Fix & modularize DataPackage fetch methods

robyngit:
Builds on top of @vchendrix changes in PR https://github.com/NCEAS/metacatui/pull/2608

### DIFF
--- a/src/js/collections/DataPackage.js
+++ b/src/js/collections/DataPackage.js
@@ -573,7 +573,6 @@ define([
       handleMemberFetchError(failedModels, errors) {
         // eslint-disable-next-line no-console
         console.error("Error fetching models", failedModels, errors);
-        console.log(errors[0].stack);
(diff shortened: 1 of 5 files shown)


robyngit:
@vchendrix I was thinking about this and wonder if we should have an AppModel setting for the fetch timeout limit per file. Or if there's not one set, treat it as no timeout. What do you think?

vchendrix:
> @vchendrix I was thinking about this and wonder if we should have an AppModel setting for the fetch timeout limit per file. Or if there's not one set, treat it as no timeout. What do you think?

@robyngit sounds good

vchendrix:
@robyngit Sorry it took so long for me to review.  I didn't see this.  Lookint at it now.

vchendrix (approved):
Works great.

PR #2618, opened 2025-02-19
Fix saving dataset after bad canonical id is set

### LINKED ISSUE #2616: Dataset editor fails to save after correcting an initially invalid canonical dataset id
After attempting to save a dataset with an invalid "canonical dataset ID", the Editor doesn't allow the user to save, even if the ID is corrected. To reproduce:

1. edit a dataset or create a new one.
2. enter an invalid canonical dataset identifier (e.g. 123)
3. press save (*important - if you don't press save here then everything works as expected*).
4. see error message ("Value URI should be an HTTP(S) URI, for example: https://doi.org/xxxx.")
5. correct identifier (e.g. https://doi.org/10.18739/A2DF6K53C)
6. press save
7. error remains and dataset cannot be saved.

I suspect this has something to do with listeners getting removed and not reset during save, but this needs to be investigated.

https://github.com/user-attachments/assets/6fd77f3e-f3f1-4cf9-a617-07f963e88d93

robyngit:
The `updateCanonicalDataset` function in the EML211 model expected the `canonicalDataset` property in the model to be an array, but sometimes it gets set as a string. When it was a string, then it was updating the annotation models with the first letter in the string (as opposed to the first item in the array). One letter is not a valid URI, which was why we were sometimes seeing an error. Corrected the function so that it handles both cases, when the `canonicalDataset` property is a string OR when it is an array.

Closes #2616

### DIFF
--- a/src/js/models/metadata/eml211/EML211.js
+++ b/src/js/models/metadata/eml211/EML211.js
@@ -174,7 +174,9 @@ define([
        */
       updateCanonicalDataset() {
         let uri = this.get("canonicalDataset");
-        uri = uri?.length ? uri[0] : null;
+        if (uri && Array.isArray(uri) && uri.length) {
+          [uri] = uri;
+        }
         let annotations = this.get("annotations");
         if (!annotations) {
           annotations = new EMLAnnotations();


robyngit:
This PR is basically a one-line change. Merging without code review!

Issue #1998, opened 2022-03-31
Use POST for all Solr queries


laurenwalker:
Right now MetacatUI only uses POST for Solr queries when the  query string is very long. I have noticed some long queries being missed, particularly when I am trying to view my profile while logged into an account that is in doens of groups. The long concatenation of usernames is too long but MetacatUI still sends it as a GET. It also is becoming somewhat tech debt to maintain both GET and POST.

We originally supported GET and POST simultaneously to give MetacatUI owners time to upgrade to [Metacat 2.10.2](https://github.com/NCEAS/metacat/releases/tag/2.10.2), which supports POSTs. That was in March 2019, so given that it is three years later, I propose we switch to sending queries via POST only. This will be a breaking change for people that want to continue using Metacat 2.10.1 or earlier, but those people can still use MetacatUI 2.20.0 or earlier.

mlhenderson:
I encountered a problem with Solr queries getting too long for ESS-DIVE after I had created quite a few teams and the data queries started to fail.  When I went through the code, I found that there is a hardcoded value in collections/SolrResults.js of 1500 for the max length of the query before it will trigger a POST if disableQueryPOSTs is false.

I was considering a PR to replace that 1500 value with a new config parameter that would default to the original value. Adjusting our deployments to permit larger GET query lengths might also be an acceptable solution.  @mbjones Do you have any suggestions?

mbjones:
I'm unsure of the logic of hardcoding a 1500 limit. Maybe it would be better to switch to using POST all the time now as Lauren had proposed in this ticket. Any downsides to that, @robyngit ?

robyngit:
@mbjones @mlhenderson 

This limit was added in 2022 to address [an issue](https://github.com/NCEAS/metacatui/issues/1712) with long GET requests failing.

Why Lauren chose 1500 specifically is unclear. Browsers and servers generally support much longer URLs than that, so it could have just been a conservative choice. The only reference I could find to 1500 specifically is the [maximum packet size for Ethernet](https://en.wikipedia.org/wiki/Maximum_transmission_unit#MTUs_for_common_media), which is 1500 bytes, so to avoid packets getting fragmented ??

I think switching to POST all the time is a good idea. The only reasons I can think of against it are:
1. It would techincally be a breaking change to change the default `disableQueryPOSTs` value from `false` to `true`, or to remove the option entirely.
[...]

PR #2642, opened 2025-03-18
Replace hardcoded value to gate Solr POST over GET with config option

mlhenderson:
Adds maxQueryLengthGETs config option to replace 1500 value in SolrResults that is used in combination with POST flag to control POST behavior for long Solr GET queries.

The default value remains the same 1500 value for backward compatibility.  Setting a low value for maxQueryLengthGETs will trigger POST for Solr queries.

This leaves it up to a deployment to control POST behavior without breaking older deployments.

Related to issue #1998


mlhenderson:
Closing in favor of a new PR that uses POST only for Solr queries.

PR #2643, opened 2025-03-18
fix: use POST for Solr queries to resolve issues with long GET requests

mlhenderson:
Fixes issue #1998

Simplifies usePOST logic in SolrResults.js to use POST for queries unless disableQueryPOSTs is true.  Docs are also updated to reflect the change.

### DIFF
--- a/docs/docs/SolrResults.html
+++ b/docs/docs/SolrResults.html
@@ -338,7 +338,7 @@
     
     <dt class="tag-source">Source:</dt>
     <dd class="tag-source"><ul class="dummy"><li>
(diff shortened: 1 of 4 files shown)


robyngit (approved):
Thank you @mlhenderson! Tested and couldn't find an issue for just using POST for all queries. Added a little change to remove a `usePOST` option used in an instance of the SolrResults model since the model isn't doing anything with that setting now.

A tip to make future contributions (🤞🏻) even easier - no need to rebuild the docs for every PR. I do that once before each release. Appreciate the thoroughness though! :)

PR #2648, opened 2025-03-24
Data subsetting

robyngit:
MVP partial download from cesium map feature


shirlysteph:
PDS testing updates from Emily Longano on March 31:

I spent Friday playing with the partial download tool. I attempted to download the IWP data set from our Prudhoe Bay watershed study area. I wanted to compare the downloaded content with the files I already have in an area I know very well. As an end user, I have a few thoughts:
- [ ] When you download a file, the name is generic "IWP". It got very confusing when I was trying to download different resolutions and different file types. It would be nice if the name could at least include the resolution or another identifying factor. I ended up downloading them one by one so I could carefully label them. My download folder looked like IWP, IWP (1), IWP (2), IWP (3)...
- [ ] 16 different resolutions is a bit overwhelming. I attempted to work with the top three resolutions. If I was on my own looking for data for WaSiM modeling for example, it would be tough to determine which resolution to download for my use. It would be beneficial to download the file in the shown resolution.
- [ ] I actually have not successfully opened a file in arcgis that I dowloaded from the metacat partial download tool. The files download as tiles. For example, the highest resolution (setting 15) yields ~100 geotiffs for our small watershed. I have not tried loading them individually as tiles into arcgis, but it would be nice to have one file instead of dozens.
- [ ] I also ran into issues on trying to load the files into arcgis. Either the file type is not supported and I get a message that the file might be corrupt or there is no location data so the small tile ends up floating in the ocean somewhere.
- [ ] @shirlysteph touched on this during the last team meeting, but in order to use the selection tool all the layers need to be selected ahead of time. It would be nice to select an area and see all the data available without having to go into the layer menu and check every single data layer to see if it exists for download in that area.

shirlysteph:
Feedback from Ben Maglio

- [ ] Maybe a future wishlist item (and no idea how difficult to code up) but it would be cool if you could upload a shapefile to use as an area of interest. Not super urgent but would be nifty.
- [ ] I'm glad you added this: "Access full downloads (including original shapefiles) in the Layers panel above." but I think it might be helpful to be more explicit, something like: "To download full datasets (including original shapefiles) please use the layers panel above"
- [ ] I agree with Emily, the resolutions tab is a little intense... I wonder if it would be better to round the numbers I.e. 4.55 -> 5, and I don't know if you need to add /px to the units. If all units are meters, you could have the units next to the drop down box. However, I think it may be nice to calculate easier to read units: e.g. 5m, 1.2km, etc.
- [ ] There is also a typo in 19567,88 with comma instead of decimal point.
- [ ] ALSO, I wonder if you can get rid of the numbering of each resolution? I think the extra number adds to the difficulty in reading.
- [ ] Small but possible annoying feature: when you select resolution and select image file format. If you change your resolution the file format gets reset... again not super urgent but it would be nice to select GeoTIFF and have that stay there as I change resolution
- [ ] I like the file size calculator very nice
- [ ] Not sure if I like the file structure once it is downloaded. I think Emily's point about the folder naming is good. Maybe some string with the layer name, resolution, and maybe also something referencing PDG: e.g. PDG_IWP_5m.png
- [ ] Additionally, a little user input box to name the file could be nice. I'm thinking a box which is automatically filled with PDG_IWP_5m, but the user can edit, say I want to call this file, "test_iwp".

Issue #1889, opened 2021-09-29
Allow downloading data directly from the cesium map


robyngit:
In the current (dev) version of the cesium map, there is a link shown in the layer details panel that leads to a configured url where a user can download the source data. We discussed the idea of allowing users to download a portion of the visible data directly from the map, for example:
- Allowing users to draw a rectangle (or other shape) around an area of interest, and download all visible data in that area
- Clicking to select multiple tiles on the map, and downloading data from those tiles.

robyngit:
Here is a quick mockup of how this feature could look:
![download](https://user-images.githubusercontent.com/26600641/201766286-8536821f-3fe5-44b3-820c-834d6d459c38.png)

mbjones:
We also could expose a WMTS URI for that download area, so that someone could access the dataset in their GIS of choice without directly downloading it all at once immediately. This would probably require a base WMTS document for a set of layers, that is then constrained to a spatial region chosen on the map.

robyngit:
We could also include a citation with each download

mbjones:
Need to consider how the user indicates which data format and what resolution of data to access from a given area of interest (AOI) or set of tiles. For example, if the view is set is at zoom level 7 for a dataset layer that includes up to zoom 15 tiles, and the user selects most of Northern Alaska, do we:

- select data tiles from zoom 7?
- select data tiles from zoom 15?
- return geopackage, geotiff, or other file formats? (PNG is unlikely to be desired)
- allow the user to select multiple data layers from those that are visible?

All of these considerations also apply when asking for a WMTS capabilities document describing the AOI.

robyngit:
Related concept: Here's a mockup from @hngrtr imagining how we could show summary statistics for given areas of interest (a pre-defined tile in this case):
<img width="1440" alt="summary-idea" src="https://github.com/NCEAS/metacatui/assets/26600641/84b88e7e-9691-45ee-8642-276044dbb50d">

mbjones:
Google doc summarizing experiments with partial data downloads in QGIS based on WMTS. Challenges with unsupported datatypes in QGIS and other libraries. See document from @yvonnesjy and @ianguerin here: https://docs.google.com/document/d/1pVGlacPihHXCqO9451bVh68VeIyHC5erPx2g0RD2UQY/edit

robyngit:
Possible MVP:
  - user selects a region of interest from the map
  - construct list of geotiff file paths (using z/x/y indices)
[...]

PR #2656, opened 2025-04-03
Fix layout issue with download panel in the cesium map

robyngit:
This PR fixes the layout issue where sometimes the download panel had a large gap between the buttons and other content. To fix it, I had to simplify the layout & css somewhat. Here are the details of the changes I made:

- The `download-panel` view was wrapped in a `draw__tool-panel` div, which added complexity because we were toggling the visibility of both divs to show/hide the download panel. Removing the outer `draw__tool-panel` div made it easier to debug the CSS issues. I could also then remove the CSS rules for the `draw__tool-panel` div and then only needed to adjust the positioning of the `download-panel` div. 

  i.e. before:

  ```html
  <div class="toolbar__content toolbar__content--active">
  <div class="draw__tool-panel" style="visibility: visible">
    <div class="download-panel" style="visibility: visible">
    ...
  ```
  after:
  
  ```html
  <div class="toolbar__content toolbar__content--active">
    <div class="download-panel">
      ...
  ```
- The draw buttons were being appending to the end of the main download panel div, but I think the intention was to add them to the `draw-tool` container. I moved them which helped simplify the CSS rules. 

- With the above changed, only the `download-panel` div needed to be toggled to show/hide the download panel. This made it possible to both simplify and move the logic for showing/hiding the download panel to the `action` property in the `section` option in the Toolbar View.

- Removed some unused imports
- Simplified some other CSS rules to avoid using `!important` unnecessarily.
- Removed some unused CSS rules
- Made the buttons grey when disabled

Related to issue #1889

### DIFF
--- a/src/css/map-view.css
+++ b/src/css/map-view.css
@@ -1410,40 +1410,6 @@ other class: .ui-slider-range */
  * Panel for drawing polygons in the map
  *
  */
-.draw__tool-panel {
-  display: flex;
-  position: absolute; /* Or relative, fixed, or sticky based on your layout */
-  justify-content: center; /* Center horizontally */
-  align-items: center; /* Center vertically */
-  bottom: 5px; /* Adjust to move it down from the top */
-  width: 100%; /* Adjust the size of the container */
-  /* height: auto; */
-  height: 80px;
-  padding: 0.5rem;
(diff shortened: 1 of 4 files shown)


shirlysteph:
Tested and looks good!

Issue #688, opened 2018-07-11
Add ability to copy attribute metadata to another entity


laurenwalker:
Design and implement a feature that allows users to copy all the attributes from one entity to another.

laurenwalker:
This work should happen on the `feature-copy-attr` branch. This feature will eventually be released with the 3.X code, but currently that branch is still on the 2.X code. I will merge into the 3.X code when that time comes.

Chris, Dominic, and I discussed this feature today and decided we would have a "Copy to..." button in the left sidebar of the entity/attribute modal window (in `EMLEntityView`). Clicking on the button will show a list of files in the right pane where users can choose one or more files to copy the attributes to. (Either checkboxes or a multi-select box). After selecting files, the user clicks an OK button and all of the attributes get copied to each of those EMLEntities.

laurenwalker:
I'm not sure if any work has been started on this, but after thinking about timelines, I think we may want to hold off on this feature until the new attribute editor is done. Open to thoughts.

dmullen17:
Sounds good to me

robyngit:
Something like this, but the copy to button will go above the attribute list, along with the "Fill from file" button when it's available

![Image](https://github.com/user-attachments/assets/c89844a8-4b88-4397-9303-e3020133da50)

<img width="538" alt="Image" src="https://github.com/user-attachments/assets/7d3743a0-e7e4-471a-8fcb-bea237a8b4aa" />

mbjones:
We already have a 'Fill from File' button. Shoudl we make the new button parallel, and name it something more like 'Copy from..." where they can then choose a documented entity to start the metadata from? The upside is this makes it parallel in operation to the Fill from file button. The downside is you would have to navigate to and click the button for each entity (rather than doing a 1-many copy from the source entity). Maybe both patterns are viable?

PR #2687, opened 2025-05-19
Enable copying attributes by reference 

robyngit:
This PR extends the attribute copying functionality (#688) to allow users to link to an existing attribute list instead of cloning it. This is supported by the EML attributeList element via <references>.

- Users can now choose between copying or linking attributes when duplicating across entities.
- This choice is available for both the "copy from" and "copy to" methods.
- The UI indicates when an entity references another attributeLis and the user can click a button to directly switch to editing the source attributeList.
- A new EML references model handles finding the linked attributeList (and can be used for other references elements in EML in the future). References are now parsed and serialized correctly in the attributeList.
- A new EMLAttributeList model handles parsing, serializing, validating, and managing a collection of EML attributes OR a references element.
- A new configurable toggle view is used to select between "Copy Attributes" and "Link Attributes."

image

image


rushirajnenuji (approved):
Thanks @robyngit for working on this  - tested by creating/editing dataset attributes on dev/stage environments and it looks great! Really improves the experience! 🎉

PR #2702, opened 2025-07-14
Add support for collapsible sections in Markdown

robyngit:
Collapsible sections are added with `<details>` and `<summary>` tags, allowing users to hide or show content as needed. To allow this, the only change needed was to the `showdown-xss-filter` configuration, which now includes `summary` as an allowed tag (`details` is already allowed).

In addition to the collapsible sections, there is now a button in the markdown editor toolbar to insert these sections easily. There is also an example in the Markdown editor example file to demonstrate how to use this feature.

Additional changes include:
- Fixing all linting errors.
- Moved the very short markdown template into the markdown view directly instead of using a separate template file and increasing the number of HTTP requests.


rushirajnenuji (approved):
Looks great, thanks @robyngit

Issue #2710, opened 2025-07-28
Use EMLUtilities.getParentEML() instead of redefining in multiple models


robyngit:
The `getParentEML` function is currently redefined in multiple places throughout models. This creates unnecessary duplication and makes maintenance harder. Now that a shared implementation exists in `EMLUtilities.getParentEML()`, we should update all code that calls the model-specific version to use `EMLUtilities.getParentEML()` instead.

PR #2711, opened 2025-07-28
Use EMLUtilities.getParentEML() in all EML models

robyngit:
Instead of redefining the same function over a dozen times, import `EMLUtilities` and use the `getParentEML` function that is defined there once. Keeps the code DRY.

I tested creating a new EML model with all fields filled and had no issues. All of the functions replaced were identical, thus I don't expect this to have any functional changes.

Issue #2710.


rushirajnenuji (approved):
Looks good to me, 👍🏻
Thank you for these changes @robyngit

Issue #1770, opened 2021-05-18
Filter map features by attribute


laurenwalker:
Support filtering features in a data layer by an attribute that exists on each feature. Design a UI widget that will show the attributes to filter on. We may also want to allow different coloring rather than filtering.

robyngit:
## Example UIs from our mockups:

<img width="369" alt="Screen Shot 2021-11-29 at 12 48 01" src="https://user-images.githubusercontent.com/26600641/143917729-fee549db-4132-4f03-91e6-ad7498f8fc5d.png">
<img width="369" alt="Screen Shot 2021-11-29 at 12 47 41" src="https://user-images.githubusercontent.com/26600641/143917732-970f63f6-c4d0-4ab0-91cb-4cad03c3c4ff.png">

robyngit:
This feature was requested by Mike Brubaker and Mike Brook for the LEO Network layer

robyngit:
Introduced in PR #2734. Still need to support multiple filters for a single layer as well as different data types and UIs (first release is a dropdown for categorical data).

PR #2712, opened 2025-07-29
Feature 1770 filter by attribute

shirlysteph:
Issue #1770

First implementation of the "**Filter by Property**" Feature. Amongst other needs, this feature is primarily built to support time-series filtering to enhance visualization of time-series datasets in the Permafrost Discovery Portal.

**Implementation:**
In order to build this feature the following steps were taken.

1. The portal configuration was updated for layers that are "filterable". The `filters` element for each property of such layers are updated to include a new sub-element called "allValues" indicating all property values by which the layer can be filtered.

3. Created the `FilterByAttributeView` to create the dropdown elements for each `FilterModel`. This view is appended to the _layer details panel_ within each _layer item_ (that are configured to be filterable).

4. The `SearchSelectView` is reused to built dropdown components for this view.

6. The `LayerItemView` is reused to toggle visibility of filtered layer features.

**Open Questions:**
1. The current name of the feature is set as "Filter by Property" to be closely aligned with the "Select By Attribute" feature found in typical GIS systems. This naming needs to be reviewed with the PDG team.
2. Naming conventions for attributes and attribute values need to be reviewed.
3. Layout of the two dropdowns may need to be adjusted.

**Next steps to improve this feature:**
1. Include toggle icon (eye icon) against each filter property in the Legend Panel. This icon visibility adjusts in tandem with the Filter by Property feature.
2. If the filter property is `Time` then add a time-slider.
3. Currently the filter functionality works for categorical property values. Extend and built this functionality to accomodate numerical filters. This might require adding a third dropdown for operators such as greater than (>), equal (=), etc.
4. Extend the `FilterByAttributeView` to allow the addition of multiple filters for each filter model that has been assigned to the layer.

### DIFF
--- a/src/css/map-view.css
+++ b/src/css/map-view.css
@@ -846,7 +846,8 @@ represents 1 unit of the given distance measurement. */
   padding: 0 1.5rem;
   cursor: pointer;
 
-  .layer-item__visibility-toggle {
+  .layer-item__visibility-toggle,
+  .layer-item__filter-icon {
     opacity: var(--map-no-brightness-or-opacity-tweaks, 0.85);
     padding: 0;
     margin: 0.25rem;
@@ -909,12 +910,6 @@ represents 1 unit of the given distance measurement. */
           );
         }
       }
-
-      .icon-filter--active {
-        fill: transparent;
-        color: var(--portal-col-neutral-4, currentColor);
-        background-color: transparent;
-      }
     }
 
     .layer-item__settings {
@@ -922,6 +917,10 @@ represents 1 unit of the given distance measurement. */
       display: flex;
       border: 1px solid var(--portal-col-primary-2);
     }
+  }
+
+  .layer-item__filter-icon--active {
+    color: var(--portal-col-primary-6);
   }
 
   .layer-item__settings {
--- a/src/js/collections/maps/VectorFilters.js
+++ b/src/js/collections/maps/VectorFilters.js
@@ -25,6 +25,14 @@ define([
       model: VectorFilter,
 
       /**
+       * Checks if there are any active filters in the collection.
+       * @returns {boolean} Returns true if there are active filters, false otherwise.
+       */
+      hasActiveFilters() {
+        return this.some((filter) => filter.get("active"));
+      },
+
+      /**
        * This function is used to determine if a feature should be shown or hidden based
        * on the current filters.
        * @param {Object} properties The properties of the feature to be filtered.
--- a/src/js/models/maps/VectorFilter.js
+++ b/src/js/models/maps/VectorFilter.js
@@ -121,15 +121,23 @@ define(["jquery", "underscore", "backbone"], function ($, _, Backbone) {
           // Store a copy of the initial value selection as a default
           this.set("defaultValues", [...modelValues]);
 
-          // If neither values nor allValues exist, then filterModelAvailable is set to false.
-          // This is used to show/hide the Filter by Property panel.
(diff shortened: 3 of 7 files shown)


src/js/views/maps/FilterByAttributeView.js
```diff
@@ -0,0 +1,285 @@
...
+
+        if (!this.filterValuesSelection) {
+          const filterValueSelect = new SearchSelect({
+            options: attributeAllValues,
+            placeholderText: "Values",
+            inputLabel: "Select value(s)",
+            allowMulti: true,
+            allowAdditions: false,
+            separatorTextOptions: false,
+            selected: defaultAttributeValues,
```
robyngit, line 213:
```suggestion
            selected: [...defaultAttributeValues],
```

wrapping an Array in a spread operator (`[... {array} ]`) makes a **duplicate** of the original array. We have to do this for the selected values that we pass to the SelectView because the SelectView modifies the array directly. Because arrays are [reference types](https://codeburst.io/explaining-value-vs-reference-in-javascript-647a975e12a0), when the SelectView modifies the array, it will modify the **original** array that's in the filter model. It does this silently, meaning there's no `"change:values"` event that gets triggered.

This can lead to unexpected behavior in the application.

When I was testing the SurfaceWater layer in the PDG map, the pre-selected value "0" wasn't showing up, *unil* I added the spread operator to make a copy of the array.

src/js/templates/maps/layer-item.html
```diff
@@ -1,12 +1,14 @@
 <button class="<%= classes.visibilityToggle %> map-view__button" title="Toggle layer visibility">
   <i class="icon-eye-open"></i>
 </button>
-<span class="<%= classes.label %>">
+
+<span class="<%= classes.label %> layer-item__label-wrapper">
```
robyngit, line 5: Rather than adding a new `layer-item__label-wrapper` class here, it would be better to modify the class that already is set on the label (whatever `classes.label` is set to). This keeps the CSS easier to read and understand. It also allows to keep using dynamic css classes here which makes the view easier to update moving forward.

src/js/views/maps/LayerDetailsView.js
```diff
@@ -118,6 +120,13 @@ define([
           showTitle: false,
           hideIfError: true,
         },
+        {
+          label: "Filter by Property",
+          view: FilterByAttributeView,
+          collapsible: false,
+          showTitle: true,
+          hideIfError: true,
```
robyngit, line 128:
```suggestion
          hideIfError: true,
          renderIf: (model) => model?.get("filters")?.length > 0,
```

We could add a conditional here to decide whether or not a section should get rendered. Then we can reuse this logic for other sections in the futurel.

src/js/views/maps/LayerDetailsView.js
```diff
@@ -206,6 +215,15 @@ define([
               detailSection.el.style.display = "none";
             }
           }
+          // Hide the Filter by Property view if there is no filter model for the selected layer
```
robyngit, line 218: Here we have to depend on the label of the section: "Filter by Property". If we change that label in the future, then we have to remember to change it here too. It makes the code fragile. To make it more robust, we could use the `renderIf` function mentioned above. Right before we create the `detailSection` we can return if the `renderIf` function returns false. e.g.:

```js
// If the section has a renderIf function, check if we should render
// this section or not
if (
  typeof section.renderIf === "function" &&
  !section.renderIf(model)
) {
  return null; // Skip rendering this section
}
```

robyngit (approved):

PR #2733, opened 2025-08-26
Fix bug with namespaces in attribute serialization

### LINKED ISSUE #2728: EML attribute serialization namespace issue
In some cases when copying attributes in the Metadata Editor, the resulting `<attribute>` elements are being serialized with an `xmlns="http://www.w3.org/1999/xhtml"` declaration, indicating that the element is in the HTML/XHTML namespace instead of the EML namespace. This causes a validation error when saving the EML document.

The issue is likely caused by mixing parsing/serialization approaches:  
- Newer code in the editor is using the native `DOMParser` / `XMLSerializer` with `application/xml`  
- Older MetacatUI code still uses jQuery's HTML-based parse/serialize methods (`$.parseHTML`, `$(...).html()`), which create nodes in the XHTML namespace  

When elements are created in one namespace and later serialized in another document, the browser adds the `xmlns` declaration to preserve their original namespace.

Haven't figured out how to reproduce it yet, but the issue arose when copying attributes from one entity to another in the editor.

Example output:
```xml
<attribute xmlns="http://www.w3.org/1999/xhtml" id="urn-uuid-1234...">
```

robyngit:
This PR fixes a bug with XML serialization of EML <attribute> elements. Most of the codebase still uses HTML-oriented parsing methods like jQuery’s parseHTML, while I've been writing newer code using modern XML APIs. The mix creates a namespace mismatch: elements parsed with parseHTML end up in the HTML namespace, and when these are inserted into an XML DOM and serialized, extra namespace declarations are added. The result is invalid EML output.

To resolve this, a new utility method was introduced to serialize DOM elements to strings while stripping out unwanted namespace attributes. This utility replaces direct calls to XMLSerializer in the two places where <attribute> elements are serialized: the EMLAttributes collection and the EMLAttributeList model.

Closes #2728

### DIFF
--- a/src/js/collections/metadata/eml/EMLAttributes.js
+++ b/src/js/collections/metadata/eml/EMLAttributes.js
@@ -5,7 +5,8 @@ define([
   "backbone",
   "models/metadata/eml211/EMLAttribute",
   "models/DataONEObject",
-], ($, Backbone, EMLAttribute, DataONEObject) => {
+  "common/EMLUtilities",
+], ($, Backbone, EMLAttribute, DataONEObject, EMLUtilities) => {
   /**
    * @class EMLAttributes
    * @classdesc A collection of EMLAttributes.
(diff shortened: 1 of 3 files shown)


rushirajnenuji (approved):
Looks good 👍🏻

Issue #2714, opened 2025-07-29
Add the ability to show/hide vector feature layers directly from the legend


shirlysteph:
Description: Add an icon of some sort that will allow the user to toggle layer visibility directly from the Legend Panel instead of doing it through the Filter by Property view. In a sense do what Filter by Property does, but as a short-cut. May need to do some mock-ups to determine the best way to do this without overcrowding the Legend Panel.

Context: This is a feature addition to the "Filter by Property" feature (see issue https://github.com/NCEAS/metacatui/issues/1770) and extension of issue #2713. The filter feature currently allows the user to toggle the visibility of vector features on the map based on filterable properties. For example, the Permafrost Zones layer can be filtered and their visibility adjusted based on various zone types. The values for some of these properties (e.g., zone types) are configured in the legend using the color palette set in the portal configuration document. The current request is to toggle visibility of layers from within the legend panel.

PR #2734, opened 2025-09-02
Update legend swatches to show/hide vector feature layers directly from the legend

shirlysteph:
**See issue https://github.com/NCEAS/metacatui/issues/2714 **

### Description: 
Adjust the look of labels in the legend, based on whether the feature layer is visible on the map or not. This adjustment is a change in label transparency. In addition, also allow the user to select (i.e., show/hide) vector layers based on pre-configured filter values directly from the legend view.

### Status: 
This UI adjustment is a first iteration of testing whether this is an ideal way to show/hide vector feature layers directly from the legend.

### Context: 
This feature is an improvement to the "Filter by Property" feature (see issue https://github.com/NCEAS/metacatui/issues/1770).

### Code changes:

**Base code:** Include code updates presented in PR ##2723.

Additionally,

1. Added the layer-item class to each swatch item in the legend. Re-using the class that was already created for layer panel for this purpose. The addition of this class to each swatch is part of the `CategoricalSwatchView.js `code.
2. Include the filter icon to indicate that layer can be filtered. This icon is added only for layers that have a filter model set on them. This code for this is in `LayerLegendView.js`.
3. The filter model is selected and set in the legend panel by matching property label with the property label of the Color Palette model. This code is part of `LegendContainerView.js`.
4. Changing the opacity and visibility of filter values is handled by the `onLegendItemClick()` function added to the `CategoricalSwatchView.js `code.
5. When all filter values are unselected in the legend view for a particular layer, the layer visibility is set to false, and the a) filter icon is turned off in the layer panel; b) the layer legend is removed from the legend panel; c) the last set of filtered values are cached in memory.
6. The attribute values dropdown in the 'Filter by property' view is updated when filter values in legend view are selected/unselected. This is handled by a new listener set on the filter model in the `VectorFilterView.js`. The function that gets triggered is `renderValueSelect()`.

### DIFF
--- a/src/css/map-view.css
+++ b/src/css/map-view.css
@@ -839,15 +839,15 @@ represents 1 unit of the given distance measurement. */
   height: min-content;
 }
 
-.layer-item {
+.list-item {
   display: grid;
   grid-template-columns: min-content auto max-content;
   align-items: center;
   padding: 0 1.5rem;
   cursor: pointer;
 
-  .layer-item__visibility-toggle,
-  .layer-item__filter-icon {
+  .list-item__visibility-toggle,
+  .list-item__filter-icon {
     opacity: var(--map-no-brightness-or-opacity-tweaks, 0.85);
     padding: 0;
     margin: 0.25rem;
@@ -861,7 +861,7 @@ represents 1 unit of the given distance measurement. */
     color: var(--portal-col-neutral-4, currentColor);
     background-color: transparent;
 
-    .layer-item__icon {
+    .list-item__icon {
       fill: var(--portal-col-neutral-4, currentColor);
       svg > path {
         fill: var(--portal-col-neutral-4, currentColor);
@@ -877,17 +877,17 @@ represents 1 unit of the given distance measurement. */
     background-color: var(--portal-col-neutral-1, #323745);
     transition: background-color 0.3s ease-in-out;
 
-    &.layer-item--shown {
+    &.list-item--shown {
       background-color: var(--portal-col-primary-1, #15324e);
     }
 
-    .layer-item__settings {
+    .list-item__settings {
       /* Show button when layer is hovered. */
       display: flex;
       border: 1px solid var(--portal-col-neutral-4);
 
       /* Use the same style on button hover and when selected. */
-      &.layer-item--selected,
+      &.list-item--selected,
       &:hover {
         background-color: var(
           --portal-col-primary-2,
@@ -897,11 +897,11 @@ represents 1 unit of the given distance measurement. */
     }
   }
 
-  &.layer-item--shown {
-    .layer-item__visibility-toggle {
+  &.list-item--shown {
+    .list-item__visibility-toggle {
       color: var(--portal-col-primary-6, var(--map-col-highlight__deprecate));
 
-      .layer-item__icon {
+      .list-item__icon {
         fill: var(--portal-col-primary-6, var(--map-col-highlight__deprecate));
         svg > path {
           fill: var(
@@ -912,24 +912,24 @@ represents 1 unit of the given distance measurement. */
       }
     }
 
-    .layer-item__settings {
+    .list-item__settings {
       /* Show button when layer is shown. */
       display: flex;
       border: 1px solid var(--portal-col-primary-2);
     }
   }
 
-  .layer-item__filter-icon--active {
+  .list-item__filter-icon--active {
     color: var(--portal-col-primary-6);
   }
 
-  .layer-item__settings {
+  .list-item__settings {
     /* By default, don't show button. */
     display: none;
     align-items: center;
     border-radius: var(--map-border-radius-small);
 
-    .layer-item__settings-toggle {
+    .list-item__settings-toggle {
       font-size: 0.85rem;
       width: 1rem;
       height: 1rem;
@@ -941,20 +941,20 @@ represents 1 unit of the given distance measurement. */
     }
 
     /* Use the same style on button hover and when selected. */
-    &.layer-item--selected,
+    &.list-item--selected,
     &:hover {
       display: flex;
       background-color: var(--portal-col-primary-1, #15324e);
       border: 1px solid var(--portal-col-primary-1, #15324e);
 
-      .layer-item__settings-toggle {
+      .list-item__settings-toggle {
         color: var(--portal-col-primary-6);
       }
     }
   }
 }
 
-.layer-item__label {
+.list-item__label {
   font-size: 0.825rem;
   font-weight: 400;
   margin: 0 0.5rem;
@@ -968,16 +968,16 @@ represents 1 unit of the given distance measurement. */
   gap: 4px;
 }
 
-.layer-item__highlighted-text {
+.list-item__highlighted-text {
   background-color: var(--portal-col-highlight-2, rgba(30, 127, 196, 0.5));
 }
 
-.layer-item__icon > svg {
+.list-item__icon > svg {
   height: 100%;
   width: auto;
 }
 
-.layer-item__visibility-toggle:hover {
+.list-item__visibility-toggle:hover {
   opacity: var(--map-no-brightness-or-opacity-tweaks, 0.5);
 }
 
@@ -2359,7 +2359,7 @@ other class: .ui-slider-range */
   align-items: center;
   gap: 0.5rem;
 
-  .layer-item__filter-icon {
+  .list-item__filter-icon {
     color: var(--portal-col-primary-6);
     /* width: 2rem;
     height: 2rem; */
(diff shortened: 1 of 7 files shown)


src/js/views/maps/legend/CategoricalSwatchView.js
```diff
@@ -52,6 +66,59 @@ define([
...
+        // Set a new array reference to trigger vector filter's change event
+        if (!newFilterValues?.length) {
+          // When all values are cleared from the attribute values dropdown, the
+          // layer visibility is set to false, and the filter icon is turned off
+          // (i.e., transparent).
+          this.layer.set("visible", false);
+        } else {
+          this.filterModel.set("values", newFilterValues.slice());
+        }
+        // Adjust opacity of legend item swatch
```
robyngit, line 118: I think what you originally had here but commented out works: `this.filterModel.set("values", newFilterValues.slice());` Is there a reason you switch to the if-else block?

Even when `newFilterValues` is an empty array (`[]`), setting the `"values"` in the filterModel to the empty array hides the layer because of listeners we set up during development of the filter feature.

src/js/views/maps/legend/LayerLegendView.js
```diff
@@ -83,16 +88,40 @@ define([
             name,
           }),
         );
+
+        // Insert the filter icon to the right of the label element text. This
+        //  icon appears for layers that are "filterable" based on their
+        //  atrributes.
+        this.swatchEl = this.el.querySelector(".layer-legend__layer-name");
```
robyngit, line 95: For consistency, to add classes to the `CLASS_NAMES` constant instead of using the string directly here

src/js/views/maps/legend/CategoricalSwatchView.js
```diff
@@ -52,6 +66,59 @@ define([
...
           this.model.getCss(),
         );
+
+        if (this.filterModel) {
+          // Add layer-item class to the view's main element
+          // This class is added to reuse functionality of the class from layer list items
+          this.el.classList.add("layer-item");
+
+          // Rebind event after DOM content is replaced
+          this.delegateEvents();
```
robyngit, line 76: No need to re-delegate. Backbone does this during render automatically. Unless this fixed a problem that I didn't come across in testing?

src/js/views/maps/legend/CategoricalSwatchView.js
```diff
@@ -52,6 +66,59 @@ define([
           "background-color",
           this.model.getCss(),
         );
+
+        if (this.filterModel) {
+          // Add layer-item class to the view's main element
+          // This class is added to reuse functionality of the class from layer list items
+          this.el.classList.add("layer-item");
```
robyngit, line 73: Yay! Love reusing existing classes rather than re-writing the same CSS rules over and over. Since the other class is named something so specific to layer elements, it does make it a little confusing since here we're using it for a palette swatch. Maybe renaming this class and all of its sub-classes (e.g. `layer-item__icon`) to `list-item` would be clearer that the class is re-used for different types of elements? Would need to test to make sure all the changes work everywhere `layer-item` is already used though. What are your thoughts?

src/js/views/maps/legend/CategoricalSwatchView.js
```diff
@@ -52,6 +66,59 @@ define([
...
+          // (i.e., transparent).
+          this.layer.set("visible", false);
+        } else {
+          this.filterModel.set("values", newFilterValues.slice());
+        }
+        // Adjust opacity of legend item swatch
+        const opacity = this.model.get("filterActive") ? 1 : 0.25;
+        this.$(`.${CLASS_NAMES.swatch}`).css("opacity", opacity);
+        this.$(".categorical-swatch__value").css("opacity", opacity);
       },
```
robyngit, line 122: Think about moving this to it's own little function, like "showFilterStatus" or something. Then you can call `showFitlerStatus()` here and in `render()` so you don't need to repeat the same code twice (makes it easier if we ever want to change how we indicate that a category is filtered out)/

src/js/views/maps/legend/LayerLegendView.js
```diff
@@ -83,16 +88,40 @@ define([
...
+        }
       },
 
       /** Fills the palette div with categorical swatches. */
       renderCategoricalPalette() {
         this.renderTemplate();
         this.model.get("colors").forEach((color) => {
-          const swatch = new CategoricalSwatchView({ model: color });
+          // if (color.get("value")) {
+          // const swatch = new CategoricalSwatchView({ model: color });
```
robyngit, line 115: You can remove these 2 commented out lines, as well as the `// }` below!

src/js/views/maps/legend/LayerLegendView.js
```diff
@@ -83,16 +88,40 @@ define([
...
+          const filterIconEl = document.createElement("span");
+          const classes = [
+            CLASS_NAMES.icon,
+            CLASS_NAMES.filterIcon,
+            CLASS_NAMES.filterIconActive,
+          ];
+          filterIconEl.classList.add(...classes);
+          filterIconEl.title = "Filter by property"; // add tooltip
+          filterIconEl.innerHTML = `<i class="icon icon-filter"></i>`;
+          this.swatchEl.prepend(filterIconEl);
```
robyngit, line 106: I think it looks better to the right, (`append`), plus it's then consistent with the placement of the filter icon in the layer list

src/css/map-view.css
```diff
@@ -2354,6 +2354,17 @@ other class: .ui-slider-range */
...
   font-weight: 500;
+
+  display: inline-flex;
+  align-items: center;
+  gap: 0.5rem;
+
+  .layer-item__filter-icon {
+    color: var(--portal-col-primary-6);
+    /* width: 2rem;
+    height: 2rem; */
```
robyngit, line 2365: you could remove these 2 lines

src/js/views/maps/legend/LegendContainerView.js
```diff
@@ -61,9 +62,49 @@ define([
...
+       * @returns {AssetColorPalette} - Return the updated color palette model
+       * @param layer - Layer item model
+       * @param colorPaletteModel - Asset color palette model
+       */
+      updateColorPalette(layer, colorPaletteModel) {
+        const colorPaletteProperty = colorPaletteModel.get("property");
+        let layerValues;
+        layer.get("filters")?.forEach((model) => {
+          if (model.get("property") === colorPaletteProperty) {
+            layerValues = model.get("values");
```
robyngit, line 94: Could you just do:
`const layerValues = this.filterModel.get("values")`?

Because you already found the `filterModel` in `render()`?

Haven't tested, but seems like it would work?

robyngit (commented):
Very excellent coding @shirlysteph!! The feature works perfectly. I couldn't find a way to break it or get any console errors. :) I added a few suggestions/questions/comments but they are all very minor. Let me know when you're ready and I'll merge it!

src/js/views/maps/legend/LayerLegendView.js
shirlysteph: addressed in commit #019e861 in #2734

src/js/views/maps/legend/CategoricalSwatchView.js
shirlysteph: Fix in commit [fc91ad8](https://github.com/NCEAS/metacatui/commit/fc91ad803d500a4de29c93ecb6af66f95cd85215)

src/js/views/maps/legend/LayerLegendView.js
shirlysteph: The reason I had this this on the left using prepend was I had a case (in the test portal) where the filter icon alone got moved to a new line due to spacing issues, while the label remained in the first line. It kind of looked odd. Please let me know if you think something like that would be okay, and then I can make this change.

src/css/map-view.css
shirlysteph: Fix in [5d0b682](https://github.com/NCEAS/metacatui/commit/5d0b682c67dfff9080ceb19e64ce6c0fe62f10f3)

src/js/views/maps/legend/CategoricalSwatchView.js
shirlysteph: Fix made!!

src/js/views/maps/legend/CategoricalSwatchView.js
shirlysteph: So the reasoning behind this if-else block does is as follows: When all values are unselected in the legend panel, 1) the layer is removed from the legend panel, 2) the visibility of the layer is turned off in the layer list panel, and 3) all filter values are cleared from the filter by property dropdown. Now when the user **turns on** the visibility of the layer using the eye-icon on the layer list panel, its corresponding legend is added to the legend panel, but everything stays unselected because the "values" are set to the empty array. Instead using the if-else block now we  reset the legend to the last selected filter value when the layer is reselected. Hope this clarifies :)

robyngit (approved):
@shirlysteph this feature works so well, it's very intuitive now to show/hide features. So cool!!

Issue #1918, opened 2021-11-22
Cannot update package with private data object


laurenwalker:
Scenario:
- Someone has changePermission permission on a metadata and resource map object, but does not have read permission on 2/3 data files in the package. This person can still click the Edit button and access the package in the editor. All data files still show up in the Editor, even though they are private. (This is a related bug - the files should not show up in the file listing, or should be grayed out somehow. They are visible because the metadata and resource map are accessible and parsed). The user edits the metadata and clicks Submit. The Submit button spins forever.

Only one file was accessible to this user in the MEtadata View (this is expected.)
![Screen Shot 2021-11-22 at 1.07.18 PM.png](https://images.zenhubusercontent.com/5d152e12c27d0c306f021863/f216aac0-4903-4382-9c25-92953b87e30a)

All three files show up in the Editor. Error message on DataONEObject:389 (`response is null`):

![Screen Shot 2021-11-22 at 1.03.49 PM.png](https://images.zenhubusercontent.com/5d152e12c27d0c306f021863/87e09a17-2906-47f4-aa6a-3f5725b643b3)

## Expected behavior:
- Editor should hide private data files
- A user that is "an owner"/changePermission on the metadata and resource map should be able to edit the package and save, even with private data files in the package.
- Perhaps consider adding anyone who is an "owner" of the resource map and metadata as an owner of all files in the package too. This is related to bulk access policy editing in #1829

amoeba:
> Editor should hide private data files

What do you think about always showing all objects in the package in the listing, regardless of access, adjusting the display of non-read-access objects to make it clear to the user what's going on and what they can do about them? My main reasoning is that I think any user with r/w/cP access on the package should have full control over the list of objects in the package. If we don't show non-readable objects, the user can't remove or replace them which I think could be frustrating.

laurenwalker:
> My main reasoning is that I think any user with r/w/cP access on the package should have full control over the list of objects in the package. If we don't show non-readable objects, the user can't remove or replace them which I think could be frustrating.

I agree that in most cases, someone who has chP access on the metadata and package should most likely have it for all of the data files, as well. But with our object-level permissions editing in the UI, it is entirely possible, so we need to accommodate those cases.

If a data object is private to a person, they will not be able to perform an update/replace on it. In spirit, I don't think they should be able to remove it either, although they *can* since that involves updating the resource map only.

Hiding the private files may be a step too far. Maybe just disabling the Replace button is good enough for now.

[...]

Issue #1916, opened 2021-11-15
Person removed from access policy on update


laurenwalker:
Rani reported a package with OPC where someone was added as an 'owner' of the metadata record but not the resource map. I tried to remove this person and add them back and the access policy messed up again. Here is the access policy before I updated it:

```xml
<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
<ns3:systemMetadata xmlns:ns2="http://ns.dataone.org/service/types/v1" xmlns:ns3="http://ns.dataone.org/service/types/v2.0">
    <serialVersion>0</serialVersion>
    <identifier>resource_map_urn:uuid:d61c20ba-1441-4a10-90ec-1bb2275ecf2a</identifier>
    <formatId>http://www.openarchives.org/ore/terms</formatId>
    <size>5317</size>
    <checksum algorithm="MD5">fea2a859d9dfea3f2139f9a955094e98</checksum>
    <submitter>http://orcid.org/0000-0003-1501-0861</submitter>
    <rightsHolder>http://orcid.org/0000-0001-7205-2123</rightsHolder>
    <accessPolicy>
        <allow>
            <subject>http://orcid.org/0000-0002-0594-2166</subject>
            <permission>read</permission>
            <permission>write</permission>
            <permission>changePermission</permission>
        </allow>
        <allow>
            <subject>http://orcid.org/0000-0001-7205-2123</subject>
            <permission>read</permission>
            <permission>write</permission>
            <permission>changePermission</permission>
        </allow>
        <allow>
            <subject>CN=opc-data-admins,DC=dataone,DC=org</subject>
            <permission>read</permission>
            <permission>write</permission>
            <permission>changePermission</permission>
        </allow>
    </accessPolicy>
    <replicationPolicy replicationAllowed="false"/>
    <obsoletes>resource_map_urn:uuid:5303f61a-dcbf-47df-86a9-4b4ff3246396</obsoletes>
    <archived>false</archived>
    <dateUploaded>2021-11-15T21:12:31.056+00:00</dateUploaded>
    <dateSysMetadataModified>2021-11-15T21:12:31.237+00:00</dateSysMetadataModified>
    <originMemberNode>urn:node:CA_OPC</originMemberNode>
    <authoritativeMemberNode>urn:node:CA_OPC</authoritativeMemberNode>
    <fileName>resource_map_urn_uuid_d61c20ba_1441_4a10_90ec_1bb2275ecf2a.rdf.xml</fileName>
[...]

PR #2759, opened 2025-10-21
test: Expose critical AccessPolicy permission validation failures

mlhenderson:
Comprehensive test suite reveals multiple permission system bugs:
- Type validation failures in AppModel configuration
- Broken permission hierarchy (changePermission without write)
- Silent failures in policy creation and broadcasting
- Integration inconsistencies across views and models

Tests document expected failures and provide roadmap for fixes.


mlhenderson:
I suspect that these issues are related to the bugs contained in the skipped tests marked EXPECTED FAILURE:

https://github.com/NCEAS/metacatui/issues/1918
https://github.com/NCEAS/metacatui/issues/1916

There do not appear to be that many issues regarding access policy bugs, but it could be that this is because most data deployed with Metacat is public.

test/js/specs/unit/collections/AccessPolicy.spec.js
```diff
@@ -0,0 +1,1519 @@
...
+
+    //-----------------------------------------------------------------------
+    // 3.8  serialize()
+    //-----------------------------------------------------------------------
+    describe("serialize()", function() {
+      it.skip("EXPECTED FAILURE: always returns an <accesspolicy> element (never null)", function() {
+        const el = state.policy.serialize();
+        expect(el && el.nodeType).to.equal(1);
+        expect(el.tagName.toLowerCase()).to.equal("accesspolicy");
+      });
```
robyngit: Returning null when a node is empty is the general serialize pattern that MetacatUI uses. We use `AccessPolicy.serialize` in `DataONEObject.serializeSysMeta`:

https://github.com/NCEAS/metacatui/blob/main/src/js/models/DataONEObject.js#L1129-L1149

We don't want to add an empty accessPolicy to the sysMeta. However, is there a reason you're thinking of that makes it better to return `<accesspolicy></accesspolicy>`?

test/js/specs/unit/collections/AccessPolicy.spec.js
```diff
@@ -0,0 +1,1519 @@
...
+    //-----------------------------------------------------------------------
+    // 3.10  Event handling – “removeMe” propagation
+    //-----------------------------------------------------------------------
+    describe("event handling", function() {
+      it.skip("EXPECTED FAILURE: removes a rule when the rule fires a 'removeMe' event", function() {
+        const rule = makeRule({ subject: "temp" });
+        state.policy.add(rule);
+        expect(state.policy.length).to.equal(1);
+        rule.trigger("removeMe");
+        expect(state.policy.length).to.equal(0);
```
robyngit:
```suggestion
        state.policy.listenToOnce(state.policy, "remove", () => {
          expect(state.policy.length).to.equal(0);
        });
        rule.trigger("removeMe");
```

The original test fails because it checks the collection length immediately after triggering "removeMe", before the **asynchronous** removal actually completes. Waiting for the "remove" event makes sure the assertion runs when the model has actually been removed. It was backbone's solution of ensuring async events had completed, back before we had Promises, await/async!

test/js/specs/unit/collections/AccessPolicy.spec.js
```diff
@@ -0,0 +1,1519 @@
...
+        });
+
+        policy.set([editedBob]);
+
+        // This should fail - we expect ALL rules to be preserved
+        expect(policy.length).to.equal(4, "All rules should be preserved");
+        expect(policy.findWhere({ subject: "uid:alice" })).to.exist;
+        expect(policy.findWhere({ subject: "uid:carol" })).to.exist;
+        expect(policy.findWhere({ subject: "public" })).to.exist;
+      });
```
robyngit: Actually, this is how we expect `collection.set` to work, see the [backbone.js docs](https://backbonejs.org/#Collection-set):

> ... and if the collection contains any models that aren't present in the list, they'll be removed

We can pass the options `{ remove: false }` to `set` to keep other models, but then we will actually have 5 rules, not 4 as your test asserts. I believe backbone will only merge the two models if the model IDs are the same (pretty sure, but verify)

test/js/specs/unit/views/EditorView.spec.js
```diff
@@ -0,0 +1,370 @@
...
+        
+        state.dataObject.set("accessPolicy", rawPolicyData);
+
+        // Verify it's raw data (this is the serialization bug)
+        var policy = state.dataObject.get("accessPolicy");
+        expect(Array.isArray(policy)).to.be.true;
+
+        // Now simulate what happens during normal reload - let automatic deserialization fail
+        var jsonData = state.dataObject.toJSON();
+        var reloadedObject = new DataONEObject(jsonData);
```
robyngit: Similar questions here, why to JSON and when/how is the dataone obj data being reloaded in the app?

robyngit (commented):
Thanks for all these tests @mlhenderson, this is fantastic!! I like the new coverage report also, very helpful! I left comments on a few of the tests. Some of the "expected failures" are actually just because of little mix ups about how backbone behaves - it's not always intuitive. And the others are questions about the rationale behind testing serializing/parsing the dataONEObject as JSON. We usually use the `serialize` method to save as XML, but I could be missing something here.

Finally, I'm having trouble running a number of the new package.json scripts. For example, I can see the test site when I run `view-tests`, and for most of the new scripts I'm seeing errors like `[PAGE error] Failed to load resource: the server responded with a status of 404 (404)`. I ran `npm install` to update my local deps, but is there something else I need to do to get these working?

Thanks again for working on this!

Issue #2757, opened 2025-10-16
Add a "watch" button to add/change notifications for datasets


robyngit:
Following the mockups [in this comment](https://github.com/NCEAS/metacatui/issues/2251#issuecomment-3399917205), create a "Watch" button and add it to the header of the dataset landing page. The watch button should open a modal that allows users to check & uncheck which types of notifications they would like to subscribe to for a given dataset. The modal will first prompt the user to login if not already. We will assume that the user account already has an email address.

This issue is blocked by:
- #1234 
- #2755

mbjones:
@robyngit I looked more closely at your mockup for notifications, and it looks great.

<img width="579" height="516" alt="Image" src="https://github.com/user-attachments/assets/3e86b432-26cc-4327-94a9-dab895c7ade0" />

I think we could make the text a bit clearer for dataset changes. How about:

> Get notified when this dataset is updated with new data, improved documentation, or error corrections

I was trying to remove some jargon (metadata) and give examples of why dataset updates occur. How does that sound to you (and others)?

robyngit:
@mbjones agree we should remove "metadata" and be more clear. Slight wording suggestion for readability:

> Be notified when new data, documentation improvements, or corrections become available.

Issue #1894, opened 2021-10-06
Publish button doesn't show up sometimes


laurenwalker:
There appears to be a race condition - possibly the same one causing issues in #1832 - where the accounts service API call hasn't finished before the 'can this person publish?' checks are performed. The accounts service info is needed since we need to check if a person is in an admin group. 

more debug details...
This was noticed for a metadata doc that is in two rmaps. The rmap was updated by me via curl but I didnt update the metadata (I was just piecing back together a package that lost data files). Not sure if this is related, but it doesn lengthen the time of page load since the view has to check which rmap is the most recent version.

PR #2761, opened 2025-10-23
Restructure dataset header for clearer information hierarchy & future extensibility

### LINKED ISSUE #1639: Edit button doesn't show up sometimes
A user noticed that the Edit button every 4-5 refreshes, the Edit button would not show up. This was seen on Safari. It was the package that also discovered this bug: https://knb.ecoinformatics.org/view/urn%3Auuid%3A6d520387-06d4-47ab-acc7-722576741986 although it's unclear whether they are related.

### LINKED ISSUE #2755: Restructure dataset header for clearer information hierarchy & future extensibility
We're preparing to add a new "Watch" button to the dataset landing page header for the notification service. Before doing so, the header needs to be reorganized so that we can make room for a new button without overcrowding and complicating the layout. This will address existing inconsistencies in spacing and alignment, as well as improve the overall visual hierarchy to distinguish between different types of buttons.

It will be important to test these changes across all themes, including external themes with lots of customization (e.g. ESS-DIVE).

The mockups below (and on [**Figma**](https://www.figma.com/proto/k3ND4RerDOiXwBgcP1HkF5/Notification-Mockup?page-id=0%3A1&node-id=27-5737&p=f&viewport=254%2C507%2C0.25&t=UoskgAL6WKtwW8Ft-1&scaling=min-zoom&content-scaling=fixed&starting-point-node-id=27%3A5737)) illustrate the proposed changes, which include:
- Move the orange "lock" and blue "canoncical version" circle icons to be *inline* with the title
- Differentiate between the metrics / assessment buttons and the "action" buttons with different styling & grouping.
   - metrics: Align download, citation, and view counts on the left without button background or border.
   - actions: Align Edit, Publish with DOI, Cite, etc. buttons to the right with button styling.
- On small screens, wrap action buttons to a second line and reduce button text where possible (e.g., "Publish with DOI" → "Publish").
- Standardize button size and spacing

# Large screen

## Current
Image

## Planned
Image

# Small screen

## Current

Image

## Planned - option 1

Image

## Planned - option 2

Image

## Planned - option 3

Image

robyngit:
Create a new `DatasetControlsView` to contain logic for rendering the metrics & cite/anaylze/edit/publish buttons for a dataset. This moves the logic out of the already large `MetadataView`. The buttons that are rendered in the `DatasetControlsView` are also now fully configurable in the `buttons` property of the view, making it easy to add the Notification Service button in the future (#2757).

The styles & layout of the buttons have also been updated to match the mockups in issue #2755, with the metrics and report buttons having a more subtle appearance compared to the "action" buttons like Cite, Edit, Publish, etc. The buttons & layout also now adapt to smaller screen sizes.

During development, I fixed an issue that kept popping up where the edit and publish buttons sometimes wouldn't appear when they should. This was due to the authorization properties on the metadata model not being reset between datasets (so the permissions from the previous dataset would persist!). This fixes #1639 and #1894. I also adjusted some of the small-screen styles of the dataset view header (citation view).

Other minor changes and fixes are documented in the commit messages.

Closes #2755

## After

image

## Before

image

## After

image

## Before

image

### DIFF
--- a/src/js/views/metadataView/DatasetControlsView.js
+++ b/src/js/views/metadataView/DatasetControlsView.js
@@ -387,7 +387,7 @@ define([
         const model = this.viewModel.get("metricsModel");
 
         // Check if the model's been fetched yet
-        if (!model.get("fetched") && !model.get("fetching")) {
+        if (!model.get("synced") && !model.get("fetching")) {
           model.fetch();
         }
 


src/js/views/metadataView/DatasetControlsView.js
```diff
@@ -0,0 +1,905 @@
...
+
+      /**
+       * Get or create the metrics model used to fetch dataset metrics.
+       * @returns {MetricsModel} The metrics model instance for this view.
+       */
+      getMetricsModel() {
+        const model = this.viewModel.get("metricsModel");
+
+        // Check if the model's been fetched yet
+        if (!model.get("fetched") && !model.get("fetching")) {
```
rushirajnenuji: We do not have a fetched flag in the metrics model
```suggestion
        if (!model.get("synced") && !model.get("fetching")) {
```

robyngit: Great catch!! Thank you!!

Issue #644, opened 2018-06-08
Lofty feature request: Provide views and tools for exploring Object versions


amoeba:
A few of us were talking on Slack last week and it came up that, at least for the data team, it's often hard to understand what has changed between package and object versions from MetacatUI alone. I could see any improvement in this to be just an admin-level tool but possibly user-facing tools as well.

Some things that could be fun to implement:

- Provide clickable "There is a previous version of this package click here to view it" on the dataset landing page. We only provide a link to the newer version now.
- Provide a custom package history view that makes it easy to navigate through a package version chain
- Show the number of previous versions on the package and/or for the individual Objects. e.g., "8 previous versions". Plus a specialized view to allow the user to quickly navigate to them.
- Provide a way to show the differences between Objects entirely from MetacatUI. This could work and look similar to GitHub's diff pane. This isn't easy to do for large data files (GitHub doesn't even do this) but could work for metadata docs or small data files. Things that'd be useful to be able to glean from this tool would be:
  - What changed  between this version of the package I'm viewing and the previous? If, say, a data object was added, we could show that. If only the EML changed, and it was just that the title was changed, we could show that. If a data object was swapped out for another, we could show that.
  - A way to implement this could be to actually write an index subprocessor that knew how to detect some of these things and we could index that information so it was fast.
  - How many bytes were added/removed from the package or object between versions

I'd guess this is super low priority so if anyone wants to migrate this out of an Issue and into another design document somewhere please do that.

robyngit:
Just a note that this came up again as a desirable tool in a conversation with @dvirlar2 today. The diff tool, in particular, would be useful for anyone viewing a dataset and for the data team. For example, if somebody lands on a data package page after clicking a DOI from a citation, and they see there is a new version of that dataset, it would be great for them to see what's been changed from the version that was cited in the paper. The diff tool would also save the datateam time when trying to determine what's been updated when a new version of a dataset is saved (e.g. is it a minor typo fix, or a new data object?)

robyngit:
Related to: https://github.com/NCEAS/metacatui/issues/1549

mbjones:
Great idea, I fully support it. A first step might be to query the version history and, on any given dataset landing page, provide a dropdown to navigate through the prior and future versions relative to the current version.

I wrote a related issue on how to quickly obtain version info in https://github.com/NCEAS/metacat/issues/1429

mbjones:
[...]

PR #2764, opened 2025-10-27
Add version navigation to Metadata View

### LINKED ISSUE #153: Provide navigation for earlier and later object versions
We currently let users know that a data package or science metadata document 'has a newer version' in the UI when someone navigates to an object mid-stream in the obsolescence chain. However, being able to choose specific version earlier or later in the version history would be useful.

Mock up a versioning navigation component that would be part of the `MetadataView` and discuss pros, cons, and implementation details.  This request came from Rachael Blake, and was seconded by Matt.

robyngit:
Add buttons to the header of the dataset landing page that allow users to navigate to previous and next versions of the dataset, when they exist. Add a placeholder "dataset versions" button that will be clickable when we have a full version history view (#644)

image

Closes #153 

⚠️ **Note** this feature branches off of the `feature-2755-dataset-header-layout` branch, PR #2761. It should only be merged after the base branch is reviewed & merged.

### DIFF
--- a/src/css/metacatui-common.css
+++ b/src/css/metacatui-common.css
@@ -9402,11 +9402,17 @@ body > #extension-is-installed {
   width: 100%;
 }
 
-.citation.header .top-info > *:not(.id) {
-  text-transform: uppercase;
+.citation.header .top-info > * {
   font-size: 13px;
-  letter-spacing: 0.02em;
   line-height: 1;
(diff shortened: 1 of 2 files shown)


rushirajnenuji (approved):
Great work @robyngit, version traversing works great! 🎉

PR #2774, opened 2025-11-17
Use ORCID API v3 in Lookup model

### LINKED ISSUE #2762: update ORCID search url
The App Model is not creating a valid ORCID search URL, which causes all orcid searches via LookupModel.orcidSearch to fail.

The app model sets up the url like so:

```js
if (this.get("orcidBaseUrl")) {
              this.set(
                "orcidSearchUrl",
                this.get("orcidBaseUrl") + "/v1.1/search/orcid-bio?q=",
              );
            }
```

then elsewhere overwrites it to

```js
 if (typeof this.get("orcidBaseUrl") != "undefined")
            this.set(
              "orcidSearchUrl",
              this.get("orcidBaseUrl") + "/search/orcid-bio?q=",
            );
```

Neither work. [Version 1.1 was deprecated in 2018](https://info.orcid.org/api-version-policy/#h-background). Requests to `https://orcid.org/search/orcid-bio?` return a 302 and no results. orcid is on version 3, which uses the url format: `https://pub.orcid.org/v3.0/search/?q=`.

We should update the URLs and test to make sure everything works as expected. We should also check whether there are request limits to the v3 api.

robyngit:
This PR updates the ORCID URLs set in the AppModel to [v3](https://info.orcid.org/orcid-api-3-0-is-here/), introduced in 2019. It was previously using a deprecated version (v1?) which no longer returned results from ORCID.

The `orcidSearch` method in `LookupModel` is the only place where the `orcidSearchUrl` is actively used. This method is used on the User Profile View where a user searches for members to add to a group. In addition to adapting `orcidSearch` to use ORCID v3, I also modernized the method to conform to our new style/linting rules and to use native JS methods (e.g. `fetch`) in place of jQuery (e.g. `$.get`). I added a unit test for the `orcidSearch` method.

One other Lookup model method, `orcidGetConcepts`, relied on the older ORCID API, but was no longer used anywhere in MetacatUI, and was therefore deprecated.

### DIFF
--- a/src/js/models/AppModel.js
+++ b/src/js/models/AppModel.js
@@ -2585,7 +2585,7 @@ define(["jquery", "underscore", "backbone"], function ($, _, Backbone) {
         // ORCID search
         const orcidBaseUrl = this.get("orcidBaseUrl");
         if (orcidBaseUrl) {
-          const searchUrl = `${NORMALIZE_URL(orcidBaseUrl)}/v3.0/search/?q=`;
+          const searchUrl = `${NORMALIZE_URL(orcidBaseUrl)}/v3.0/expanded-search/?q=`;
           this.set("orcidSearchUrl", searchUrl);
         }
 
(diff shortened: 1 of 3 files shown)


rushirajnenuji (approved):
Thanks @robyngit, calls to ORCID v3 API work as expected. 🚀

PR #2776, opened 2025-11-18
Fix parsing & serializing <source> elements in EML model

### LINKED ISSUE #2497: EML corrupted when edited in MetacatUI
- This EML (https://cerp-sfwmd.dataone.org/metacat/d1/mn/v2/object/doi:10.25497/D7P88F) was edited in MetacatUI
- The resulting data package could not be saved due to an XML error:
  image001
- The Tomcat log showed the error:
```
catalina.out-[2024-08-12 20:21:08] [info] Validation for eml is https://eml.ecoinformatics.org/eml-2.2.0

catalina.out-[2024-08-12 20:21:08] [info] [Fatal Error] :294:4: The element type "source" must be terminated by the matching end-tag "</source>".

catalina.out-[2024-08-12 20:21:08] [info] The element type "source" must be terminated by the matching end-tag "</source>".
```
- In the drafts folder, it looks like the `<source>` tags were missing their closing tags: [`RECOVER_MAP 3.1.3.3 Coastal Gradients of Flow, Sal (3)-1.xml.txt`](https://github.com/user-attachments/files/16605866/RECOVER_MAP.3.1.3.3.Coastal.Gradients.of.Flow.Sal.3.-1.xml.txt)
- The error occurred when trying to add edit access to this for another user to the package
- CERP is running MetacatUI version `2.30.0`

robyngit:
The editor uses `$.parseHTML()` rather than `$.parseXML()` (or the native XML DOMParser) to parse EML. This is for historical reasons and to maintain compatibility with the rest of the codebase.

In HTML, [`<source>` is a void element](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/source), meaning it can't contain text content and must not have a closing tag. Consequently the `<source>...</source>` elements in measurement scale get mis-parsed by the EML211 model. The closing tag is ignored and the contained text ends up outside the element node, leading malformed EML.

This issue doesn't come up often in practice because `<source>` elements can't be added via the UI, but it can happen when users create and upload EML via other tools, then try to edit it in the Editor.

The EML model already had an existing `cleanUpXML()` method that was meant to replace `<source>` elements with `<sourced>` and the rest of the code base expected this behaviour (e.g. looked for the text content of `<sourced>` elements to get the `source` text for the model). However, the replacement wasn't assigned back to `xmlString`, so it had no effect. It was also not set to replace all occurrences, only the first. Additionally, we did not revert the changes when saving back to the server, so the saved EML would contain `<sourced>` elements instead of `<source>`.

This PR fixes the `cleanUpXML()` method to properly replace all occurrences of `<source>` and `</source>` with `<sourced>` and `</sourced>`, and adds a `revertSourcedToSource()` method to revert these changes before saving the EML back to the server. I also added a unit test to verify this behaviour, with multiple `<source>` elements.


rushirajnenuji (approved):
Looks good 👍🏻

PR #2790, opened 2026-01-13
Show skipped MDQ checks in Assessment page

### LINKED ISSUE #2765: Display skipped checks for quality reports in a different section
**Describe the bug**
`result/status` in a run document can have values of `SUCCESS` `FAILURE` `ERROR` or `SKIP`. Currently, if a check has a `SKIP` status, the check is shown under the informational checks section. I'd like it to be in it's own section (possibly totally separate from the current list, but if it's easiest to just add a row that is ok too). It might also be nice to have ERROR in it's own section as well, right now I think that gets added to FAILURE, and ERROR isn't typically a researchers fault so I don't love it there.

**To Reproduce**

For an example see [this dataset's](https://test.arcticdata.io/quality/s=data-suite-0.1.0/urn%3Auuid%3A3ce72fe6-07af-4f51-84d3-361dc4e674c1).

**Expected behavior**

Minimum requested changes: 1 additional row in the dropdown list for skipped checks, 1 additional row for errored checks (only show if present)

Would be nice: A separate smaller display from the main table of results. Not sure what would look good, but since this information is less important it would be nice to de-emphasize it in whatever way we can.

**Screenshots**

Image

Image

(not sure if bug or enhancement, so adding both labels)

robyngit:
Hide skipped informational and skipped checks in the pie chart. Total checks in chart is PASS + WARN + FAIL only.

image

Fixes #2765

### DIFF
--- a/src/js/views/MdqRunView.js
+++ b/src/js/views/MdqRunView.js
@@ -89,7 +89,7 @@ define([
       status: "GREY",
       icon: "chevron-sign-right",
       buildTitle: ({ count }) =>
-        `Skipped ${count} ${PLURAL(count)}${count ? " that are not applicable." : "."}`,
+        `Skipped ${count} ${PLURAL(count)}${count ? ` that ${count === 1 ? "is" : "are"} not applicable.` : "."}`
     },
   ];
 
--- a/src/js/views/MdqRunView.js
+++ b/src/js/views/MdqRunView.js
@@ -89,7 +89,7 @@ define([
       status: "GREY",
       icon: "chevron-sign-right",
       buildTitle: ({ count }) =>
-        `Skipped ${count} ${PLURAL(count)}${count ? ` that ${count === 1 ? "is" : "are"} not applicable.` : "."}`
+        `Skipped ${count} ${PLURAL(count)}${count ? ` that ${count === 1 ? "is" : "are"} not applicable.` : "."}`,
     },
   ];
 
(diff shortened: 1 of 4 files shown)


src/js/views/MdqRunView.js
```diff
@@ -743,27 +753,27 @@ define([
...
-            perc: groupedResults.GREEN.length / results.length,
+            count: greenCount,
+            perc: greenCount / dataCount,
           },
           {
             label: "Warn",
-            count: groupedResults.ORANGE.length,
-            perc: groupedResults.ORANGE.length / results.length,
+            count: yellowCount,
+            perc: yellowCount / results.length,
```
rushirajnenuji: Should these (`yellowCount` and `redCount`) be using the same denominator as `greenCount`? i.e. `dataCount` instead of `results.length`

rushirajnenuji (approved):
Thanks @robyngit , just have a small query regarding the percentage values for Warn and Fail checks, otherwise the code looks good. 👍🏻

Feel free to merge this if my above comment is not an issue. ✅

src/js/views/MdqRunView.js
robyngit: Yes good catch! Fixed.

src/js/views/MdqRunView.js
```diff
@@ -743,27 +753,27 @@ define([
...
+        const dataCount = greenCount + yellowCount + redCount;
+
         const data = [
           {
             label: "Pass",
-            count: groupedResults.GREEN.length,
-            perc: groupedResults.GREEN.length / results.length,
+            count: greenCount,
+            perc: greenCount / dataCount,
           },
```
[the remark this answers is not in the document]
robyngit: Fixed.

PR #2795, opened 2026-02-12
Refactor version tracking cache to SysMeta-based storage and add DataONE service layer

### LINKED ISSUE #2783: VersionTracker produces incorrect version chains from cache
`VersionTracker` builds obsolescence chains from SysMeta and stores them in the browser using localforage plus an in-memory cache. It tries to persist full version information for every PID in the chain to speed up retrieval and avoid repeated DataONE API calls. In practice, storing the entire chain per PID is redundant and error-prone. When a new version is found, every record in the chain needs to be updated to reflect the new predecessor/successor relationships. Using two forms of storage (in-memory + localforage) adds complexity and increases the chance of inconsistency.

In testing, I'm seeing intermittent incorrect chains (e.g., missing versions). This suggests cached records are not being updated reliably when new versions are found or added. There also may be a concurrency issue where multiple writes to IndexedDB (via localforage) can overwrite each other, leading to partial or lost data.

These version storage issues could be the underlying cause of #2750.

A cleaner approach is to remove the in-memory cache and rely solely on localforage. The stored information persists across reloads and can be shared across tabs. Reads from IndexDB are fast enough for this purpose. We should also only store the immediate predecessor and successor for each PID instead of persisting the full chain. That keeps each record authoritative. The full chain can be reconstructed on the fly when needed.

Since this storage logic needs to handle concurrent writes, TTL/expiration, and general consistency, it makes sense to split it into a dedicated storage class rather than keeping it inside `VersionTracker`.

robyngit:
This PR refactors `VersionTracker` and `ResourceMapResolver` to improve robustness, fix version chain caching issues, and extract common  logic into reusable classes.

Previously, `VersionTracker` stored full version chains, per pid, in both memory and `localForage`. This made updates difficult when new versions appeared and could result in stale or partial chains being cached, causing all sorts of issues in the Editor and elsewhere.

Instead, with this refactor, we now:
- Cache only SysMeta per PID (in memory and in local storage, both optional).
- Build version chains at read time using obsoletes / obsoletedBy relationships.
- Centralize HTTP, retry, auth, deduplication, and caching logic into independent classes that can be reused across the app.

As part of this refactor, key behaviors that are used in both `VersionTracker` & `ResourceMapResolver`, and throughout the app, have been extracted into reusable services, including:
- Retrying requests after network errors
- Deduplicating in-flight requests (to avoid, e.g. multiple requests to download the same SysMeta before the previous has completed)
- Normalizing and validating request options
- Handling common HTTP errors
- Caching with expiration (memory + persistent storage)

For example, `VersionTracker` now relies on SysMetaService, which uses the new components to manage sysMeta retrieval, cache sysMeta XML, and retry failed requests. These services are designed to be reused elsewhere in MetacatUI (but are currently only used in `VersionTracker` and `ResourceMapResolver`).

Note that although we're using modern ES6 class syntax for new classes, and using import/export, they are still designed to be imported with require. The also add Backbone.Events as a property to allow triggering events that Backbone views can listen to.

Here is a diagram outlining the relationships between the new classes and `VersionTracker`/`ResourceMapResolver`

```mermaid
classDiagram
  class ResourceMapResolver
  class VersionTracker
  class SysMetaService
  class DataONEService
  class DataONEHttpClient
  class HttpRetryPolicy
  class DataONEHttpError
  class UrlBuilder
  class PersistentStorage
  class SysMeta

  ResourceMapResolver --> VersionTracker
  ResourceMapResolver --> PersistentStorage

  VersionTracker --> SysMetaService
  SysMetaService --> SysMeta
  SysMetaService --|> DataONEService

  DataONEService --> DataONEHttpClient
  DataONEService --> PersistentStorage

  DataONEHttpClient --> HttpRetryPolicy
  DataONEHttpClient --> DataONEHttpError
  DataONEHttpClient --> UrlBuilder
```

## New classes

1. `PersistentStorage`: shared storage layer
- Wraps localforage with TTL, `expiresAt`, optional memory cache, and per-key locks (to prevent multiple concurrent writes to the same key).
- Uses a singleton pattern with instance keys generated from the constructor configuration options so that services care share the same storage instance when configured the same way (e.g. we can have a instance for sysMeta caching across the entire app that uses the same metaServiceUrl, etc.).

2. DataONE service classes in `(models/dataONEServices)`:
- `DataONEHttpClient`: normalized requests, allows request timeouts, retries on failures, makes requests abortable, and deduplicates in-flight requests. Uses a singleton pattern with instance keys generated from the constructor configuration options, like `PersistentStorage`. Itended to be extended for specific DataONE services.
- `HttpRetryPolicy`: defines which errors to retry and how many times.
- `DataONEHttpError`: normalized error object for HTTP errors, with status code, message, etc.
- `DataONEService`: Handles token management and creates a separate cache for each logged in user vs. public data. Provides higher level methods for DataONE API calls (e.g. getSysMeta) that use HttpClient and caching.
- `SysMetaService`: fetch/upload/invalidate SysMeta using the new services, an extension of `DataONEService`.
- `UrlBuilder`: safe path encoding and URL building.

## Updated classes

1. `VersionTracker`
- Removed attempts to persist entire version chains for each PID.
- Traverses versions by fetching cached or fresh SysMeta for each PID and following obsoletes / obsoletedBy links.
- Always re-fetches the first and last version in the chain, regardless of cache state, to ensure we have the most up-to-date chain.

2. `SysMeta`
- Now focused on parsing/serialization.
- Fetch logic moved to `SysMetaService`.

3. `ResourceMapResolver`
- Uses `PersistentStorage` instead of direct localforage.
- Improves multi-RM resolution and sysmeta walk behaviour.
- Improves Solr query error handling.
- Fixes event forwarding and listener cleanup behavior.

## Affected Views

`VersionNavigationView.js` and `EML211EditorView.js` updated to align with changes to `VersionTracker` and `ResourceMapResolver`.

## Tests
Added tests for:
- `PersistentStorage`
- `DataONEHttpClient`
- `DataONEHttpError`
- `HttpRetryPolicy`
- `DataONEService`
- `SysMetaService`
- `UrlBuilder`

Updates to existing tests for:
- `VersionTracker`
- `ResourceMapResolver`
- `SysMeta`


rushirajnenuji (approved):
Thanks @robyngit, lots of great changes in this PR. I was able to run the test suite without any issues and the app works as expected based on my local testing.

The new service layer classes are clean and well designed. Makes it much more maintainable going forward. 🙌🏻

Thanks again for your time and efforts, and for detailed in code and PR documentation. LGTM! ✅🚀

PR #2797, opened 2026-02-17
Display full version history for data package

### LINKED ISSUE #2766: Expose version history of datasets
Add a tool for exploring all versions of a dataset. The tool should use the versionTracker class to build a version history of the metadata object by traversing the sysMeta.

robyngit:
---

⚠️ **NOTE:** This PR branches off of #2795. That PR should be reviewed, updated if needed, and merged to develop before this one.

---

This PR adds the `VersionHistoryView` which renders a complete version history for a given dataset. It finds the version history downloading sysMeta, inspecting the obsoleted/obsoletedBy fields, and then continuing to walk the version history until it reaches the newest and oldest versions.

VersionHistoryView

If versions are private or (rarely) are 404s, it will display appropriate messages to the user.

Version History with 401

The versionHistory view can be accessed by clicking the "Version History" link at the top of a dataset landing page, or by navigating directly to `version-history/:pid`.

Since under the hood it uses `PersistentStorage`, the sysMeta for each version is cached, so the version history should load faster after the first load. The tail-ends of the version chain are always re-fetched, though, to make sure that the view is up-to-date in case new versions have been added (e.g. new version published, or version chain repaired).

An area of improvement for a future release may be to add pagination to the history chain, for example, only fetch the first 20 versions, then require user interaction to continue searching for more versions. However, I've tested datasets with >50 versions and it doesn't feel too slow. Because the view displays versions as they are found, it feels pretty responsive. Any in-flight sysMeta requests are also aborted when the user navigates away.

---

Adds feature from issue #2766


robyngit:
Note, working on adding a few small changes based on team feedback:
- [x] Show error messages before results in the status banner at the top of the timeline
- [x] Make sure dates are consistently using the same timezone, show timezone, and use `yyyy-mm-dd` format for version dates.
- [x] Highlight DOI versions, maybe add ability to filter/hide non-DOI versions
- [x] Show relative date in <older> <newer> badges, e.g. <2 days older>, <3 mins newer>
    - note: this means that we need the "latest" and "first" badges to be separate from the relative date badges. In other words, the oldest version will have two badges, one indicated how old it is relative to the version we're looking at, and another badge indicating that it's the first in the version chain.
    - [x] could be useful to add tooltips to the badges for more context
- [x] Detect and show message when we've reached max number of versions (200) in either direction (newer or older). This is the "max jumps" property we set on VersionTracker. We can let a user know that they can look at the versionHistory for the oldest or newest version to see more versions.
- [ ] Detect every obsolescence anomaly where the obsoleting PID has an earlier `dateUploaded` than the obsoleted PID (manual update by datateam), then annotate the affected younger version row and show a page-level message.
   - Consider switching to sort by obsolescence chain order rather than by `dateUploaded` property. We can still group by date for contiguous chain segments. i.e. the same date can appear in multiple "buckets".

rushirajnenuji (approved):
This looks great Robyn, thank you! A very helpful addition to MetacatUI! 🚀

Issue #2661, opened 2025-04-15
Refactor DownloadPanelView to use ExpansionPanelView for individual layer sections


robyngit:
The `DownloadPanelView` currently creates a list of expandable UI sections manually for each selected layer:

<img width="479" alt="Image" src="https://github.com/user-attachments/assets/9d9acaa7-ecd8-434c-9273-7fffe209ca5d" />

These sections mimic the structure and class names of the [`ExpansionPanelView`](https://github.com/NCEAS/metacatui/blob/develop/src/js/views/maps/ExpansionPanelView.js), but the implementation is duplicated and slightly divergent. For example, classes like `download-expansion-panel__toggle` and `download-expansion-panel__content` were adapted from `ExpansionPanelView`, but the component itself is not being used.

Instead of duplicating styles and manually managing open/close behaviour, we should refactor `DownloadPanelView` to use the actual `ExpansionPanelView` for each individual layer section.

Rebuilding a version of `ExpansionPanelView` inside `DownloadPanelView` creates extra maintenance overhead and makes it harder to apply consistent styling. Using `ExpansionPanelView` will align the `DownloadPanelView` with the MetacatUI patterns and reduce the need for custom CSS.

To see how `ExpansionPanelView` is used in other parts of the codebase, see:
  - [`LayerCategoryListView`](https://github.com/NCEAS/metacatui/blob/develop/src/js/views/maps/LayerCategoryListView.js) (each layer is expandable)
  - [`ViewfinderView`](https://github.com/NCEAS/metacatui/blob/develop/src/js/views/maps/viewfinder/ViewfinderView.js) (map views are collapsible under the "Zoom to..." section)

### Tasks

- [ ] Create a small Backbone View to encapsulate the layer section content (currently rendered inline in `generatePreviewPanel`), e.g. `LayerDownloadView`
  - This view should handle the dropdowns, checkboxes, and any other interactive elements for the layer
  - It should also manage its own state (e.g., whether the dropdown is open or closed)
- [ ] For each selected layer, instantiate an `ExpansionPanelView` with:
  - `title`: The layer's display name
  - `icon`: Optional — can reuse or omit depending on visual design
  - `contentViewInstance`: The new Backbone View that handles dropdowns, checkboxes, etc.
- [ ] Replace the existing DOM construction logic in `generatePreviewPanel` with instantiation and rendering of `ExpansionPanelView` instances
[...]

PR #2814, opened 2026-04-27
Feature 2661 Create Layer Download View

DarianGill:
## What
This PR addresses #2661 by creating a new view LayerDownloadView.js to encapsulate each of the data product options that the appear for selected layers in the DownloadPanelView.js. See the issue for a further discussion of design choices against using ExpansionPanelView.js.

## Why
**Benefits**: 

- LayerDownloadView gives us better separation of concerns, as DownloadPanelView doesn't need to own per-layer DOM construction, event wiring, or expand/collapse logic. 
- It also allows us to pull the layerSelection() and fileTypeSelection() state out of the DOP and into JS properties on tracked view instances. Previously they were read from global DOM traversal where they could in theory match elements outside the panel. The new array of views is better scoped.
- The CSS is now more maintainable/readable as it's been rewritten from a nested duplicative structure (copied and modified from ExpansionPanelView) to a flat BEM structure with a modifier class for the status warnings

**Costs**:
- There's bidirectional coupling (but at least now it's more explicit?). LayerDownloadView calls back into downloadPanelView.layerSelection(), fileTypeSelection(), getRawFileSize(), updateTextbox().
- We're re-implementing expand/collapse behavior instead of re-using that provided by ExpansionPanelView.

## How

- Extracted logic from the DOM construction loop in DownloadPanelView.generatePreviewPanel() to its own Backbone view LayerDownloadView.js
- Added checkbox state and expandability to LayerDownloadView.js and coupled them with the checkbox change event
- Replaced the duplicated .download-expansion-panel css block of nested css rules with 4 flat BEM classes: .layer-download__header, .layer-download__checkbox, .layer-download__title, & .layer-download__content and control expansion behavior with .show_content class
- Updated DownloadPanelView to use the new LayerDownload View, replacing document.querySelectorAll DOM reads with reading layerDownloadViews[i].isSelected / .selectedFileType directly.
- Added .downloads-textbox--warning CSS modifier with a ::before pseudo-element for the ⚠ prefix to make
 the download status warnings more notable so the user wouldn't get confused about what was included in a download
- Fixed a resolution change bug where changing one row's resolution dropdown overwrote it for every other row and didn't disable the download button even though the filetype dropdown would be reset
- Updated downloadSizeLimit to be 1,000,000,000 bytes to match the 1GB comment instead of the 1,050,000 that put it at about a Mb
- Added null guards for png and wmts file types in the dropdown like we had for tiff and geopackages
- Linked the input labels to the file and resolution dropdowns so clicking them focuses the right dropdown
- Removed CSS border around .download-data-list and .download-data-list__panel options that was distracting during empty state on light and dark themes

## Testing and Documentation

- `npm run test` passes without warnings or errors
- Manual testing on local host with production src/config.js maintains prexsiting behaviors (and some prexisting bugs)
- Added jsdocs since 0.0.0 and with a screenshot to LayerDownloadView.js


src/js/views/maps/DownloadPanelView.js
```diff
@@ -88,7 +88,7 @@ define([
        * this size, the download will be blocked.
        * @type {number}
        */
-      downloadSizeLimit: 1050000, // 1 GB
+      downloadSizeLimit: 1000000000, // 1 GB
```
robyngit, line 93: It's good we switched this to bytes, but I see now that other parts of the `DownloadPanelView` are using KB. For example, the `fileSizes` property uses `the file size for tiled data in geotiff format (in KB)`. `getRawFileSize()` uses these KB constants to calculate the estimated total download size. The JSdocs for the fn say that it returns `bytes`, but in fact it is returning KBs. `updateTextbox()` converts that value to MB with `/1000`. So with `downloadSizeLimit = 1000000000`, the guard allows roughly 1TB of estimated data instead of 1GB.

One fix would be to switch all the constants and calculates to bytes, for consistency, then use [`Utilities.bytesToSize()`](https://github.com/NCEAS/metacatui/blob/bd7aebfeb34f4a72bfb168453366c8a56961c0c3/src/js/common/Utilities.js#L223-L248) for display purposes.

src/js/views/maps/LayerDownloadView.js
```diff
@@ -0,0 +1,313 @@
...
+          fileSizeInfoBox.classList.add("downloads-textbox--warning");
+          fileSizeInfoBox.classList.remove("error", "wmts-copy");
+          // Only notify the parent if a file type was previously selected —
+          // that's the only case where the save-button state can change (from
+          // enabled back to deactivated). If no file type was selected before,
+          // the save button is already deactivated and no update is needed.
+          if (hadFileTypeSelected) {
+            downloadPanelView.layerSelection();
+          }
+        });
```
robyngit: I was able to download a layer even when the UI showed "Select resolution and file format to download". If another selected layer has a valid resolution and file type, the Download button stays enabled, and the incomplete layer gets downloaded using its previously selected resolution/file type.

This seems to happen because when the resolution is changed, `selectedFileType` is cleared in the UI, but the existing entry in `dataDownloadLinks[item.layerID]` is not removed. So the internal state still contains the old values, and `downloadData()` includes that stale entry. The resolution change handler should clear that entry or follow the same logic used when the file type is reset.

```suggestion
        // Resolution change: enable file-type dropdown, clear its value, update hint
        resolutionDropdown.addEventListener("change", () => {
          view.selectedResolution = resolutionDropdown.value;
          fileTypeDropdown.disabled = false;
          fileTypeDropdown.value = defaultFileTypeOption.value;
          view.selectedFileType = "";

          delete downloadPanelView.dataDownloadLinks[item.layerID];

          fileSizeInfoBox.textContent = "Select file format to download...";
          fileSizeInfoBox.classList.add("downloads-textbox--warning");
          fileSizeInfoBox.classList.remove("error", "wmts-copy");
          downloadPanelView.layerSelection();
        });
```

I removed the `hadFileTypeSelected` conditional in the suggestion because even if the save button state does not always change, calling `layerSelection()` here keeps the parent state synchronized after the queued entry is deleted.

robyngit (changes requested):
@DarianGill reviewing your PR in detail just confirms what I said in our last meeting, this is awesome! Agree that using the expansion panel here is not actually the ideal solution like I thought when I wrote the original issue. Encapsulating the per-layer logic in a view and re-using the expansion panel styles is better. Good choice.

Everything looks great, except for two little issues I found. Let me know if you want to discuss, otherwise I think it's good to merge after those are fixed!

src/js/views/maps/LayerDownloadView.js
DarianGill: Oop! Good catch. Thank you!

robyngit (approved):

src/js/views/maps/DownloadPanelView.js
DarianGill: Sticking with Utilities.bytesToSize() as it's the repo convention even though som users may expect KB instead of KiB.

PR #2822, opened 2026-05-06
Feature 2663 use Backbone events() in DownloadPanelView and LayerDownloadView

### LINKED ISSUE #2663: Replace addEventListener with Backbone events() in DownloadPanelView
The `DownloadPanelView` currently uses `addEventListener` directly in several places to bind DOM events. While this works, it’s not the best practice in Backbone apps. Instead, we should use [Backbone's built-in event handling `events()` property](https://backbonejs.org/#View-events).

Using `events()` automatically handles unbinding events when the view is removed: Backbone removes events when `view.remove()` is called. With `addEventListener`, you have to manually track and remove listeners — otherwise, you risk memory leaks and unexpected behaviour. Using `events()` also keeps event logic scoped to the view and easier to manage.

At the same time, we can make sure to use the `CLASS_NAMES` constant for all class names in the `events()` hash.

This refactoring will be easier to complete *after* we've moved all of the class names into the `CLASS_NAMES` constant (#2662). This will ensure that we have a single source of truth for class names and can easily update them in one place if needed.

### Example

Current approach:
```js
const closeDownloadPanelButton = this.el.querySelector(
  ".download-panel-close__button",
);
closeDownloadPanelButton.addEventListener("click", () => {
  view.close();
});
```

Refactored approach:

```js
events() {
  const CN = CLASS_NAMES;
  return {
    [`click .${CN.closeButton}`]: "close",
  };
}
```

Note: This assumes the `CLASS_NAMES.closeButton = "download-panel-close__button"`

## Tasks
- [x] Identify all uses of `addEventListener` in `DownloadPanelView` and determine if they can be replaced with entries in the `events`
- [x] Add missing class names to `CLASS_NAMES` to enable consistent selector usage (if not already complete)
- [x] Move those handlers into the `events()` method using dynamic keys (e.g., `click .${CN.someClass}`)
- [x] Remove the corresponding `addEventListener` calls from methods like `renderToolbar()`
- [x] Verify all handlers still function correctly (especially buttons for draw, clear, save)

DarianGill:
## Notice: ☝️
**This feature branches from work proposed in #2818. Please review and merge those changes prior to evaluating this one, or point the PR at the feature-2662-DownloadPanelView-BEM branch to see the meaningful diff before merging to develop.** 

## What:
The DownloadPanelView and LayerDownloaView used to use addEventListener directly in several places to bind DOM events. While this worked, it’s not the best practice in Backbone apps. Instead, this PR updates them to use [Backbone's built-in event handling events() property](https://backbonejs.org/#View-events) thereby closing #2663.

## Why:
Backbone's delegated events hash (or at minimum this.listenTo) is the idiomatic approach. Manual addEventListener calls are not cleaned up when undelegateEvents() or remove() is called, causing memory/listener leaks if the view is re-rendered.

## How (brought to you by Claude sonnet 4.6):

**DownloadPanelView**

- events() — added two new entries: 
  1. click .download-panel__close-button → "close" 
  2. click .layer-download__copy-icon → "handleCopyIconClick".
- handleCopyIconClick(event) — new method. Uses event.currentTarget.closest(...) to find the WMTS info box and its text node, then writes to the clipboard. Replaces the old per-render .onclick assignment entirely.
- renderToolbar() — removed the querySelector + addEventListener("click", ...) block for the close button. The delegated events() entry handles it now.
- updateTextbox() — removed the post-render DOM queries (wmtsText, copyIcon), the guard if (!wmtsText || !copyIcon) return, and the entire copyIcon.onclick = ... closure. Also removed the now-unused const view = this alias and replaced remaining view. references with this..

**LayerDownloadView**

- events() — new method returning three delegated change entries: 
  1. checkbox → handleCheckboxChange
  2. resolution dropdown → handleResolutionChange
  3. file-type dropdown → handleFileTypeChange.
- extracted three new named handler methods from the old addEventListener closures (handleCheckboxChange(), handleResolutionChange(), handleFileTypeChange()) now using this.* instead of captured local variables.
- render() — removed const view = this, removed the entire // ── Event listeners section and its three addEventListener blocks.

## Testing and Documentation:
All 1303 tests completed and passed and new handlers have comments describing them above.

### DIFF
--- a/src/js/views/maps/DownloadPanelView.js
+++ b/src/js/views/maps/DownloadPanelView.js
@@ -142,6 +142,8 @@ define([
         const events = {};
         const CN = CLASS_NAMES;
         events[`click .${CN.button}`] = "handleButtonClick";
+        events[`click .${CN.closeButton}`] = "close";
+        events[`click .${CN.copyIcon}`] = "handleCopyIconClick";
         return events;
       },
 
@@ -468,6 +470,35 @@ define([
       },
 
       /**
+       * Handles a click on the copy icon in the WMTS info box. Copies the WMTS
+       * URL text to the clipboard and briefly shows a confirmation message.
+       * Because this handler is registered via the delegated events hash, it
+       * replaces the previous `.onclick` assignment and never stacks.
+       * @param {Event} event - The click event.
+       */
+      handleCopyIconClick(event) {
+        const infoBox = event.currentTarget.closest(`.${CLASS_NAMES.wmtsCopy}`);
+        if (!infoBox) return;
+        const wmtsText = infoBox.querySelector(`.${CLASS_NAMES.wmtsText}`);
+        if (!wmtsText) return;
+        const text = wmtsText.textContent;
+        navigator.clipboard
+          .writeText(text)
+          .then(() => {
+            wmtsText.textContent = "Copied to clipboard!";
+            setTimeout(() => {
+              wmtsText.textContent = text;
+            }, 2000);
+          })
+          .catch(() => {
+            wmtsText.textContent = "Copy failed!";
+            setTimeout(() => {
+              wmtsText.textContent = text;
+            }, 2000);
+          });
+      },
+
+      /**
        * Toggles the draw tool on and off.
        * @param {boolean} [draw] - If true, the draw tool will be turned on. If
        * false, it will be turned off. If not provided, the draw tool will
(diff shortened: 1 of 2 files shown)


robyngit (approved):
Looks good to me. Tested the behaviour in the prod PDG portal and delegated events behaviour is preserved. Much cleaner to have the handlers in their own methods. 👍🏻

PR #2829, opened 2026-05-14
Feature 2665 Show total download progress & partial failures in download panel progress bar

### LINKED ISSUE #2665: Improve progress bar logic to reflect total download progress and partial failures
The current progress bar in `DownloadPanelView` is updated within a loop that processes each selected layer individually. As a result, the progress bar reflects only the progress of a single layer at a time and resets or jumps multiple times throughout the download process. This makes it hard for users to understand how far along the **total** download process is.

In addition, if one or more layers fail to download (e.g., due to size limits, 404 errors, etc.), this is not clearly communicated to the user. The progress bar may show "Download Complete!" even if some layers failed.

Instead, we want to implement a more cohesive progress tracking system that shows a combined progress bar that reflects total progress across all selected layers and tiles. For this we need a way to track the number of successful vs failed layers and report partial success/failure clearly. Ideally, we'd like to show a per-layer status in the UI (e.g., "Layer X: ✓", "Layer Y: failed")

## Problem 1: Per-layer loop causes confusing progress UI

Right now, in `downloadData()`, we loop over each layer in `dataDownloadLinks` and initiate its download individually, like this:

```js
Object.entries(this.dataDownloadLinks).forEach(async ([layerID, data]) => {
  // download logic...
  updateStatusBar(progress);
});
```
This means:
- Each layer download reports its own progress (e.g. 0–100%)
- The progress bar is reset or overwritten on every iteration
- You never see overall progress for the full batch — just per-layer spikes that overwrite each other

So, if you’re downloading 4 layers, you might see the bar jump to 30%, then reset to 0%, then jump again — it's not cumulative or meaningful at the overall level.

## Problem 2: Failures are not visible or clearly tracked

If one layer fails (e.g. 404, size too large), but others succeed:
- The final progress bar may say "Download complete!" even if some layers failed
- The user doesn’t get a clear indicator that only some layers succeeded
- Errors are shown temporarily and then overwritten by the next success message

## Proposed Solution

Here is a rough idea of what we want and how we can do it, but it may need some more thinking to fully work out on how to implement it.

- Track total number of data layers and total number of tiles across all layers
- Update the progress bar based on total tiles downloaded, not per-layer progress
  - e.g., 10 tiles total → update bar every tile download (1/10, 2/10, etc.) regardless of which layer the tile comes from
- Store per-layer status (success or error) during download
- After all downloads finish, summarize results: e.g.
  - "All layers downloaded successfully"
  - "2 of 3 layers downloaded. 1 failed: Layer XYZ (Download size exceeded limit)"
- Update `updateStatusBar()` logic to handle this final summary
  - Make sure errors do not get overwritten by later updates
- Consider modifying `retrieveDataFromURL()` to report progress by individual tile rather than per-layer
- We may need a centralized progress tracker Model that counts all pending fetches
- We could consider moving the progress tracker to its own View

DarianGill:
## Notice: ☝️
This feature branches from work proposed in https://github.com/NCEAS/metacatui/pull/2818, https://github.com/NCEAS/metacatui/pull/2822, and #2823. Please review and merge those changes prior to evaluating this PR, or point the PR at the `feature-2664-DownloadPanelView-best-practices` branch to see the meaningful diff before merging to develop.

## What:
Closes https://github.com/NCEAS/metacatui/issues/2665 by: reporting tile download progress across all layers instead of reseting after each selected layer finishes downloading, reporting partial failures in a summary instead of showing the last layer's downloaded state, and managing button state and downloads to prevent re-entry bugs and ongoing background downloads. Also renames the "Save" button to the "Download" button to better describe its purpose.

## Why:
To make the progress bar and download process more honest and reliable as well as easier to maintain and extend in future.

### Progress bar — full rewrite

The previous implementation ran downloads sequentially using `forEach` + `async` (which doesn't await), with per-layer byte-based progress tracked inside a `then`-chain. It had no aggregate view across layers.

**The new implementation:**
- Downloads all eligible layers **in parallel** using `Promise.allSettled`
- Tracks a single **aggregate tile-completion counter** shared across all layers, driving one progress bar
- Moves the progress container from being created imperatively in JS to **declared in the template** (`download-panel__progress-track` wrapping `download-panel__progress-bar`, plus a `download-panel__progress-error` div below)
- **Instantly resets** the bar to 0% (no animation) at the start of each new download, avoiding the shrink-then-grow visual
- **Separates error state** from progress state: on failure the bar switches to a red `--no-data` variant with the summary inline, and per-layer error details appear below with a ⚠️ prefix
image
image
image

### Download failure UX

Previously failures produced no visible indication in the progress bar. Now:
- A summary (e.g. *"0 of 1 layers downloaded (1 failed)"*) appears in a red error bar
- Each failed layer's reason appears below the bar

### Button state reliability

- Introduced `canDownloadAnyLayer()` as a single source of truth for whether the save button should be enabled, replacing three separate inconsistent checks in `layerSelection`, `fileTypeSelection`, and `updateLayerInfoEl`
- Save and Clear buttons are **disabled for the duration of a download** to prevent re-entrant calls
- Buttons are correctly re-enabled when the download settles, with the save button state re-derived from `canDownloadAnyLayer()` rather than assumed

### Cancellation

- `downloadData` creates an `AbortController` stored as `this.downloadAbortController`
- `reset()` calls `abort()` before teardown, cancelling all in-flight tile and metadata `fetch` calls immediately
- The post-`allSettled` path checks `signal.aborted` and restores default button state without relying on `reset()` having done it

### Dead code removed

- Removed unused `MESSAGES` entries: `downloadComplete`, `downloadFailed`, `downloading`, `generatingZip`, `fileSizeExceedsLimit`, `metadataError`, `progress`
- Removed unused instance properties: `mode`, `mapinteraction`
- Removed the imperative progress bar DOM creation from `generatePreviewPanel`

### Minor fixes

- `draw: false` declared on the prototype (was previously only set inside `toggleDraw`)
- Progress bar `border-radius` moved to the track wrapper so the fill doesn't have rounded ends that clip incorrectly at partial widths
- Progress container width fixed to `calc(100% - 60px)` to align with the layer list

## Testing and Documentation:
All 1303 tests completed and passed and new functions have comments describing them above.


DarianGill:
Also, I'm not sure why this is, and it predates this PR (it's happening in production), but downloading a selection of Ice Wedge Polygons always returns a `No data available for selected data layer(s) within area of interest` error when downloading as a geopackage, but succeeds in the other file formats even when the selected polygon doesn't change and certainly contains features.

mbjones:
Is it possible the configuration for the download tool is not pointing at the correct URI path for that product?

DarianGill:
@mbjones sounds plausible. This is what adding console prints to getRawFilesize() gets me:

```
} else if (fileFormat === "gpkg") {
  [baseURL] = gpkgDownloadLink.split(this.tileMatrixSet);
  console.log("[gpkg] raw endpoint:", gpkgDownloadLink);
  console.log("[gpkg] derived baseURL:", baseURL);
  console.log("[gpkg] sample URL:", `${baseURL}${this.tileMatrixSet}/${resolution}/0/0.gpkg`);
}
```

Raw endpoint: https://arcticdata.io/data/10.18739/A2KW57K57/iwp_geopackage_high/WGS1984Quad/15/{TileCol}/{TileRow}.gpkg
Derived baseURL: 
https://arcticdata.io/data/10.18739/A2KW57K57/iwp_geopackage_high/
Sample URL: 
https://arcticdata.io/data/10.18739/A2KW57K57/iwp_geopackage_high/WGS1984Quad/8/0/0.gpkg

Iʻve created a new [issue #2830](https://github.com/NCEAS/metacatui/issues/2830) where we can discuss/diagnose further later.

src/js/views/maps/DownloadPanelView.js
```diff
@@ -1615,17 +1708,24 @@ define([
           return { fileName, blob };
         });
 
-        // Wait for all fetches to complete
-        const results = await Promise.all(fetchPromises);
+        // Use allSettled so an AbortError from a cancelled signal does not
+        // suppress blobs that were already fully fetched in this batch.
+        const results = await Promise.allSettled(fetchPromises);
```
robyngit, line 1762: What is the rationale here for continuing with zipping and saving the completed tiles if the user explicitly cancels the download?

I understand using `allSettled()` for normal partial failures, like missing or failed tiles, so we can still ZIP the successful tiles and show warnings. 

If the abort came from the user clearing or closing the panel, in what case would they still want to download partial ZIP afterward?

Would it make sense to keep `allSettled()` for partial tile failures, but check `signal.aborted` after `retrieveDataFromURL()` and before ZIP generation/clicking the link?

src/js/views/maps/DownloadPanelView.js
```diff
@@ -1328,167 +1322,250 @@ define([
...
+            .flatMap((r) => r.value?.warnings ?? []),
+        ];
+        const failed = results.length - succeeded + skippedLayers.length;
+
+        if (failed === 0) {
+          view.updateStatusBar({
+            progress: 100,
+            message: MESSAGES.downloadSummary(succeeded, layers.length),
+            errorDetails,
+          });
```
robyngit, line 1496:
```suggestion
        const failed = results.length - succeeded + skippedLayers.length;
        const total = layers.length + skippedLayers.length;

        if (failed === 0) {
          view.updateStatusBar({
            progress: 100,
            message: MESSAGES.downloadSummary(succeeded, total),
            errorDetails,
          });
```

`layers.length` only includes attempted layers, not skipped ones.

robyngit (changes requested):
It works now! Tested with very large downloads and downloads that included failures (where the resolution + format combo has no data), and it is so much clearer now. 

Just one small change needed where the total count in the user-facing message doesn't include the skipped layers, otherwise I only have a few questions and/or suggestions.

First, `reset()` aborts an in-flight download. Should we also abort in `remove()` too?

When all but one layer succeeds, the status messages are red and look like errors. I wonder if a yellow warning would be more reflective of some success, some error?

src/js/views/maps/DownloadPanelView.js
DarianGill: Oop you're right, thanks for the catch!

src/js/views/maps/DownloadPanelView.js
DarianGill: ☝️😧 ✊😐 ...there is no rationale. That makes a lot of sense to me! I think I missed the cancellation case when updating to allSettled().

DarianGill:
I really appreciate you catching those things. I've added a yellow warning state to the progress bar and detail message for partial failures and reserved the red for if everything fails. Ready for another look!

robyngit (approved):
Great! 🎉

PR #2851, opened 2026-07-14
Feature 2635 url encode active viewfinder

### LINKED ISSUE #2635: Cesium share URL should encode the current viewfinder view if active
The "share url" functionality in the viewfinder encodes the current layers and location so that a user can return to the same view. It would be helpful to also encode the viewfinder view that is active so that it remains highlighted when a user returns.

### LINKED ISSUE #2850: Enabled layers aren't reliably encoded in url
**Describe the bug**
While testing I noticed toggling swi, Cdlbns, Ciwp, Cdarts_viz (and other) layer's visibility in the layer menu wouldn't immediately update the url. Instead they'd only appear there when I toggled another layer's (like Ciwp-coverage) visibility. 

**To Reproduce**

1. Go to https://arcticdata.io/catalog/portals/permafrost
2. Click on Layers to open the layer menu
3. Turn off all layers except the osm basemap
4. Turn on the Ice Wedge Polygon layer
5. Refresh the page
6. Note that Ice Wedge Polygon layer is now toggled off and only osm basemap is visible

**Expected behavior**
The ice wedge polygon layer is added to the el query param in the url when toggled on and restores its visibility when the page is refreshed. 

**Screenshots**
Before IWP
Image
After turning on IWP
Image
After Refresh
Image

- OS: macOS
- Browser: Chrome

DarianGill:
## What:
This PR extends the map share URL restore state so it can preserve the active viewfinder action and the open map panel, not just the camera position and selected layers. It introduces schema-based URL state with `sv`, `a`, and `op`, and moves layer-visibility restoration into the map lifecycle so shared links consistently reopen the same view, thus closing #2635.

Along the way, I also found a [bug](https://github.com/NCEAS/metacatui/issues/2850) that led to certain layers not persisting in the deployed el parameter, and patched it by making a flattened centralized layer model that's passed as a reference instead of copied, and moving the layer change listener from individual map assets to map asset categories (0409065589d996869065bb74b042778d0def96a7). I also updated the behavior of layer parsing to hide any layers that don't have `visible` or `configfuredVisibility` set because they could slip through as undefined previously and updated the el parameter to encode the edge case of no layers enabled instead of being omitted when nothing is toggled on.

## Why:
The existing share URL returns users to the same map extent and layer set, but it loses the active viewfinder context. That means shared links do not keep the selected viewfinder card/button highlighted, and they do not reliably reopen the relevant panel or visualization state. Adding explicit restore-state fields makes the shared URL behave more like a true permalink to the user’s current map experience.

## How: 
Brought to you by GPT 5.4mini ...

### File Edits:
- `SearchParams.js` replaces the older destination/layer helpers with a normalized restore-state API that parses and writes `sv`, `a`, `op`, destination, enabled layers, and future schema fields while preserving unrelated query params.
- `Map.js` now owns map restore-state parsing and application, including layer-visibility normalization, configured-vs-runtime visibility handling, and URL-driven restoration during initialization..
- `MapAssets.js` preserves existing `MapAsset` model instances when flattening assets so shared state such as visibility and selection remains intact.
- `AssetCategory.js` stops reading layer visibility directly from URL params and instead respects `configuredVisibility` versus runtime `visible` state when building category assets.
- `MapAsset.js` removes the older per-layer URL sync listener logic that was updating enabled layers directly from asset models.
- `ViewfinderCardModel.js` normalizes stable action IDs for viewfinder buttons from the config so actions can be restored from the URL.
- `CesiumWidgetView.js` no longer owns search-param sync; it now focuses on camera publishing, consumes restore state through the map flow, syncs enabled layers from live layer groups, and cleans up Cesium listeners correctly.
- `MapView.js` parses restore state once, stores it on the map model for child views, and reopens the visualization panel when an active visualization URL is already present.
- `ToolbarView.js` is URL-agnostic and event-driven: it emits open-panel changes, restores the initial panel from model state, and no longer manages search params directly..
- `ViewfinderCardView.js` reports which action was activated, passes the full action into active-state handling, and adds programmatic activation by of actions that re-applies visual active state without simulating DOM clicks in restore flows.
- `ViewfinderCardsListView.js` threads action-activation callbacks from individual cards up to the list view so the parent viewfinder can sync URL state. It also coordinates per-list active-card reset so only the currently activated card/button remains visually active inside the list
- `ViewfinderCards.js` now parses leonetwork configs fetched from a url into updated viewfinder card configs with button actions and replaces the /thumbnail/ images with higher res /resized/ ones--falling back to the thumbnails if the image doesn't load.
- `ViewfinderView.js` now owns active-action restore orchestration at the viewfinder level: it reads restore state, locates the rendered action by ID across categories/lists, opens the card category's expansion panel (without simulating clicks), and asks the card view to restore that action. It synchronizes active action ids back into URL state on activation, and clears active action URL state when visualization closes. 
- `map-view.css` now assigns a border and box transition for viewfinder cards and a more appropriate (longer and darker) box shadow when active.

### Testing:
- `SearchParams.spec.js` updates coverage for the new restore-state API, including schema parsing, partial destination parsing, explicit empty layer state, unrelated query param preservation, and clearing known restore keys.
- `Map.spec.js` now carries most restore logic validation that previously sat closer to views: URL-driven layer restoration for flat and categorized layers, explicit no-layers-selected behavior, default-hidden semantics for unspecified visibility, configuredVisibility normalization, and fail-fast rejection of preconstructed MapAsset instances in normalization inputs.
- `Map-Search.spec.js` updates expectations for map-to-search-param synchronization after restore-state ownership shifted further into the map model lifecycle.
- `MapView.spec.js` adds coverage ensuring the visualization panel reopens when a visualization URL is already active.
- `AssetCategories.spec.js` verifies that flattened asset collections reuse the same `MapAsset` instances rather than cloning them.
- `AssetCategory.spec.js` removes tests tied to the older direct URL-driven visibility behavior in `AssetCategory`.
- `CesiumWidgetView.spec.js` was simplified to match the refactor that removed direct SearchParams coupling from CesiumWidgetView; tests now focus on Cesium widget responsibilities (including listener lifecycle/cleanup) while URL restore assertions were moved to Map specs.
- `ViewfinderCardView.spec.js` adds coverage for preserving explicitly configured action IDs, and adds coverage for action activation behavior, restore-by-action flows, and active styling restoration without synthetic DOM clicking.
- `ViewfinderView.spec.js` adds restore-flow coverage for finding active actions by id, reopening the containing category/panel, and reapplying active state after render.
- `ViewfinderModel.spec.js` updates assertions impacted by the allLayers accessor/getter refactor and related viewfinder selection wiring.
- `SearchView.spec.js` and `LegendContainerView.spec.js` received follow-up expectation updates tied to shared map-layer access changes, not new restore-state surface area.

## Testing:
Updated test specs to replace old url encoding helpers with new ones, and enforce new id generation, panel re-openening, visibility restoration, map asset reuse (instead of copying), and url parsing behavior.
- Ran `npm run test`
- Result: `1396` passes, `0` failures

## Documentation:
All new functions have '@since 0.0.0' and the new URL restore-state contract is documented in code with the new search param constants and restore-state helpers. I also documented the enabled layers bug I discovered in https://github.com/NCEAS/metacatui/issues/2850 and marked it as closed by this PR.


src/js/models/maps/viewfinder/ViewfinderCardModel.js
```diff
@@ -115,6 +230,15 @@ define(["underscore", "backbone", "models/maps/GeoPoint"], (
...
         const ids = layerIds ?? [];
+        const cardSeed = [
+          rest.title || "",
+          rest.description || "",
+          rest.image || "",
+          lat ?? "",
+          lng ?? "",
+          alt ?? "",
+          ids.join(","),
+        ].join("|");
```
robyngit: This nearly duplicates `buildCardSeed()`. It also looks like `ViewfinderCardModel.parse()` isn't used yet (instead parsing the cards actually happens in the parent collection, but that's a different problem). If we continue needing the `cardSeed`, we could move it out of `parse` since Backbone calls `initialize()` after the parsed attributes have been set, so this would give us a single ID-generation path.

src/js/models/maps/Map.js
```diff
@@ -319,12 +359,16 @@ define([
...
+            const normalizedLayers = config.layers.map(normalizeLayerVisibility);
+            const layers = new MapAssets(
+              normalizedLayers,
+            );
             this.set("layers", layers);
             this.get("layers").setMapModel(this);
             this.unset("layerCategories");
             this.set("allLayers", layers);
           }
+          this.restoreLayerVisibilityFromUrl();
```
robyngit: I think visibility needs to be normalized once, before any `MapAsset` models are constructed.

Right now flat layers are normalized here in `Map.js`, categorized layers are normalized separately in `AssetCategory`, and the URL override is applied after construction below, in `restoreLayerVisibilityFromUrl()`. That ordering matters because the concrete Cesium assets call `createCesiumModelWhenVisible()` during initialization. A visible configured visible omitted from `el` can start loading before the later override hides it.

Could we parse the URL state once in `Map.initialize()` and run both `config.layers` and each category’s `layers` through the same helper before creating the collections? The helper would preserve the portal value in `configuredVisibility`, then set the initial runtime `visible` value from `el` when it is present. `AssetCategory` could then pass its prepared configs directly to `MapAssets`, and the post-construction restore call would no longer be needed during initialization.

src/js/views/maps/CesiumWidgetView.js
```diff
@@ -767,22 +785,46 @@ define([
...
-          }
+      /**
+       * Get enabled layer ids from live layer groups for URL state sync.
+       * @returns {string[]} A normalized list of visible layer ids.
+       * @since 0.0.0
+       */
+      getEnabledLayerIdsForUrlState() {
+        const layers = this.model
+          .getLayerGroups()
+          .flatMap((group) => group?.models || []);
```
robyngit: Are we doing this because `this.model.get("allLayers")` is a snapshot that gets stale? I'm wondering if we should set `allLayers` at all, and instead just add a map method `map.getAllLayers()` that returns a flat list that is always up to date.

src/js/views/maps/CesiumWidgetView.js
```diff
@@ -213,8 +213,12 @@ define([
...
 
-          const destination = SearchParams.getDestination();
-          if (this.model.get("showShareUrl") && destination) {
+          const destination = SearchParams.parseStateFromUrl().destination;
+          const hasCompleteDestination =
+            destination.latitude != null &&
+            destination.longitude != null &&
+            destination.height != null;
+          if (this.model.get("showShareUrl") && hasCompleteDestination) {
             // Go to position specified in query params.
```
robyngit: Since the `SearchParams` logic is getting more complicated now, I wonder if we should take the opportunity to try to get `SearchParams` out of `CesiumWidget` view altogether? The original design plan (which I see I never documented anywhere, sorry!) was to keep as much logic as possible outside of `CesiumWidgetView` such that it's easier to replace Cesium with some other mapping tool in the future.

Here, for example, we could parse the URL directly in the Map model and update the `zoomTarget` in the `MapInteraction` model. Otherwise, should we want to say replace Cesium with MapBox, we'd have to essentially copy paste this SearchParam --> map position logic in MapBox. Really, all a map widget has to know is that it should navigate to a set of coords. It doesn't need to know or care about where those coords originate from.

I totally acknowledge that the mixed responsibilities existed here before you started working on this, so it completely makes sense you added it here. Just thinking it might be a good time to intervene and keep the CeisumWidgetView from taking on even more functionality than needed.

src/js/views/maps/CesiumWidgetView.js
```diff
@@ -737,7 +755,7 @@ define([
         const events = Cesium.ScreenSpaceEventType;
 
         // Remove previous listeners if they exist.
-        view.removeMouseListeners;
+        view.removeMouseListeners();
```
robyngit, line 735: Oh my! Good catch!!!

src/js/views/maps/CesiumWidgetView.js
```diff
@@ -1897,7 +1939,12 @@ define([
       onClose() {
         this.destroy3DTilesInspector();
         this.removeMouseListeners();
+        this.removeCameraListeners();
         this.removeNavigationListeners();
+        if (this.removePreRenderLightListener) {
+          this.removePreRenderLightListener();
+          this.removePreRenderLightListener = null;
+        }
```
robyngit, line 1879: Also a great catch, thank you!

src/js/views/maps/ToolbarView.js
```diff
@@ -514,6 +523,66 @@ define([
...
+      },
+
+      /**
+       * Sync the current open panel selection to URL restore state.
+       * @since 0.0.0
+       */
+      syncOpenPanelInUrl() {
+        if (!this.shouldSyncUrlState()) return;
+        const openPanel = this.getActiveRestorePanelId();
+        SearchParams.updateOpenPanel(openPanel);
```
robyngit: Tell me what you think, but wondering if it's worth making all of the active open panel logic more general, and moving the `SearchParam` dependency out of `ToolbarView`? Instead the panel that should be open initially could be passed directly to this view. The view could then update the Map model directly or emit an event anytime the active panel changes?

src/js/views/maps/viewfinder/ViewfinderView.js
```diff
@@ -80,11 +84,94 @@ define([
...
+                  categoryCid: listView.categoryCid,
+                }
+              : null;
+          }, null);
+        }, null);
+
+        if (match) return match;
+
+        return null;
+      },
```
robyngit, line 142: If I'm following, the flow is:
1. Start with the stable `actionId`.
2. Find the actual action.
3. Get the action index.
4. Pass the index into `activateActionByIndex()`
5. Look up the action from the index again.

Why not return the action itself?

src/js/models/maps/Map.js
```diff
@@ -26,6 +28,44 @@ define([
...
   }
 
+  /**
+   * Normalize configured and runtime visibility for a layer model/object.
+   * configuredVisibility tracks the portal-configured value, while visible
+   * tracks the current runtime value (which may be overridden from URL state).
+   *
+   * If both values are missing in config, default to hidden.
+   * @param {Backbone.Model|object} layer A layer config object or model.
+   * @returns {Backbone.Model|object} The normalized layer.
```
robyngit: I don't think we can safely support preconstructed `MapAsset` instances here. They may have started loading already, and their defaults make it impossible to tell whether visibility was omitted from the original config. Since the map configuration is documented as an array of `MapAssetConfig` objects, I'd remove that model instance branch rather than imply it provides the same behaviour.

src/js/views/maps/CesiumWidgetView.js
robyngit: Wondering if we should also add a general `view.stopListening`, e.g. to remove the asset listeners

src/js/views/maps/viewfinder/ViewfinderCardView.js
```diff
@@ -198,6 +201,28 @@ define([
...
+
+        if (btn.classList.contains(CLASS_NAMES.buttonSecondary)) {
+          this.selectLayers({ currentTarget: btn });
+          return true;
+        }
+
+        this.handleButtonClick({ currentTarget: btn });
+        return true;
+      },
+
```
robyngit, line 261: This restores the previous action by emulating a user click, which does more than restore the card's active state. We don't need all of those side effects during restoration. The map position and enabled layers are already restored from their own URL parameters, so running a map action again can overwrite that newer state. I think opening an external tab automatically when the page loads would be unexpected as well.

To show how this could be confusing, try the following:

0. Set config to `demo.adc`
1. Go to: http://localhost:3000/portals/pdg_embed_vis/
2. Click on "View Layers" button in Virtual Tours
3. Click the Reset button in the toolbar to return to the pan arctic overview
4. Go to the layers panel and hide the Bing Satellite layer and show the OpenTopoMap layer
5. Refresh the page
6. See that the page returns to the Virtual Tour point with the Bing Satellite layer still visible and the OpenTopoMap layer hidden

I think restoration should handle each part explicitly:

- Mark the matching card and button as active.
- Reopen an iframe action because that visualization is part of the saved in-app view.
- Let the existing camera and layer parameters restore map actions.
- Never automatically reopen a tab action. It can remain highlighted if we want to remember the last selected action, but it should only open following an actual user click.

The second part of this is that we should clear the active action when a user changes the map position or toggles layers other than those that are configured for the `view layers` action.

robyngit: Adding to this, it might be useful to ensure we scroll an active action into view when a user restores their page. Otherwise, it's really not obvious that a certain action is active when you come back to the page and it's hidden at the bottom of the viewfinder list.

Alternatively, we could consider making cards expand to fill the entire viewfinder panel when active. This would make it more obvious that an action is still active and also provide a way to show more details about the view like in Doug's original design.

This might be a follow up issue.

src/js/models/maps/viewfinder/ViewfinderCardModel.js
```diff
@@ -5,6 +5,103 @@ define(["underscore", "backbone", "models/maps/GeoPoint"], (
...
+      attrs.image || "",
+      latitude ?? "",
+      longitude ?? "",
+      height ?? "",
+      Array.isArray(attrs.enabledLayerIds) ? attrs.enabledLayerIds.join(",") : "",
+      Array.isArray(attrs.enabledLayerLabels)
+        ? attrs.enabledLayerLabels.join(",")
+        : "",
+    ].join("|");
+  };
```
robyngit: When we were implementing the functionality that saves the currently visible layers in the URL, we had discussion about how to identify layers. We did consider using the label as the ID, but decided against it since it's a mutable property. We ended up deciding that if a layer does not have an explicit ID configured, then it simply won't be encoded in the URL. Given that all of these properties can be updated, I'm not sure we should rely on a fallback ID. What do you think?

robyngit (commented):
Thanks for taking this on. There is a lot of excellent work here and this is a good direction! The consolidated URL state is easier to follow than the older parameter-by-parameter handling, and the cleanup around the map listeners fixes either problems we didn't know we had or problems that will become evident down the line!

I'm requesting changes because I think a few parts of restoration could be improved. Details are in the comments, but the main point is to interpret the saved state once, apply it before dependent models start loading, and avoid restoring state by emulating a user click.

src/js/models/maps/viewfinder/ViewfinderCardModel.js
DarianGill: Oof you're right, seems like the collection parse also absorbed some card model responsibilities over time and theres redudancy in the model. 

In f1fb7e9ddfccb0f4d6ff6e12bb0d8d52de04bdb0 I refactored the normalization into helpers in ViewfinderCardModel.js that are shared with ViewfinderCards.js and kept the collection parse for response-level branching for the various input shapes (url config, plain card array, leo network geojson).

Now:

- ViewfinderCardModel.js owns card-local normalization through a shared normalizeCardAttributes helper. That helper resolves legacy position versus top-level lat/lng/height, applies default map-button presentation, creates GeoPoint, and only synthesizes a legacy map action when there is not already an explicit map action.
- ViewfinderCardModel parse is now just a thin wrapper around that shared helper.
- Action ids are generated in initialize from parsed this.attributes using buildCardSeed, so both direct model parse and collection-created models use the same id-generation path.
- ViewfinderCards.js keeps only collection responsibilities: response-shape branching, LEO handling, and map-layer enrichment such as enabledLayerIds, enabledLayerLabels, and featureLayer.

I also added two focused regression tests in ViewfinderCardView.spec.js: one for fallback ids on parsed legacy map actions, and one to ensure an explicit map action does not get a duplicate synthesized legacy map action.

src/js/models/maps/Map.js
DarianGill: Good idea. I got turned around implementing this after catching the enabled layer bugs and didn't think it through. I attempted to implement this in a9b391f41406a806ba02430f56d44e4089e0206d by adding new helpers to map.js and using them to parse visibility once in initialization before any map models are created for both flat and categorized configs. 

Here's the implementation:
1. Added pre-construction visibility helpers in Map.js:32:
  - URL state parser: parseLayerVisibilityStateFromUrl Map.js:36 
  - URL override resolver: getUrlVisibilityOverride Map.js:49
  - Unified per-layer normalizer (preserves configuredVisibility, sets runtime visible from el when present): normalizeLayerVisibility Map.js:68
  - List/category normalizers used for both shapes: 
    - normalizeLayerListVisibility Map.js:114 - maps normalizeLayerVisibility over a  list
    - normalizeLayerCategoryVisibility Map.js:125 - unpacks categories then calls normalizeLayerListVisibility and reassembles them
2. Updated Map initialization ordering in Map.js:433:
  - Parse URL visibility once at start of initialize: Map.js:436
  - Normalize category layers before AssetCategories construction: Map.js:438
  - Normalize flat layers before MapAssets construction: Map.js:448
  - Removed the old post-construction restore call during initialize (so no late flip after Cesium asset init path).
3. Removed duplicate category-local normalization from AssetCategory.js:70:
  - AssetCategory now directly consumes prepared layer configs: AssetCategory.js:75
4. Removed restoreLayerVisibilityFromURL() Map.js:566 since it was only called there and is no longer necessary 

If that's too fragmented/bespoke we could probably generalize with just normalizeLayerVisibility and a normalizeLayerGroupVisibility function that works for both lists and categories.

src/js/views/maps/CesiumWidgetView.js
DarianGill: Yeah that was the idea. I think allLayers is set once during map init and not updated after. I didn't see the actual mechanism that had for making visibility stale, though there is a TODO (Map.js:460) noting that reordering or adding/removing layers would make it stale. I updated all layers to reference the active layer models since that seemed to be part of what was contributing to the el restore bug, so in theory it should actually represent the active visibilty state now and never be stale. I think you're right though that it may make more sense to just implement a map.getAllLayers, especially if we're planning to support adding/removing/reordering layers in future. It could just flatten the object returned by getLayerGroups Map.js:586.

allLayers is currently consumed by LegendContainerView, DownloadPanelView, ViewfinderModel and ViewfinderCards, but I don't see why they couldn't call map.getAllLayers() instead.

DarianGill: I did that in f3b65662545997ab2707d6263e5be12f8850af2f and replaced all the tests that expected allLayers as well.

src/js/views/maps/CesiumWidgetView.js
DarianGill: I like separating the responsibilities. Attempted to do this in b947ff135875151bd6ea71fe448ecf20a843ef64 and update the corresponding tests in 8640b274ff2d2806f42b21aff6bfe270e218c15d.

The map model now owns URL restore/sync, and Cesium just publishes camera state through MapInteraction.cameraPosition and consumes zoomTarget.
- Map.js now parses restore state, applies the initial destination via zoomTo(), and writes URL state from shared interaction/layer state.
- MapInteraction.js now stores cameraPosition, which is the widget-agnostic camera contract the model can observe.
- CesiumWidgetView.js no longer imports or calls SearchParams; it only updates camera state and navigates to zoomTarget.
- MapView.js no longer parses restore state itself.

src/js/views/maps/CesiumWidgetView.js
DarianGill: Done in beb8d089c73cabcc02de386b064071f8c62f5b89

src/js/views/maps/ToolbarView.js
DarianGill: Yeah that sounds like a good separation of concerns in line with pulling search params out of the CesiumWidgetView. I implemented it in e983afda3e0eae52afbe0ffe5611510a6d607cd3. 

Got rid of SearchParams panel constants and replaced them with local panel ids in ToolBarView. Also added an initial panel input, initialOpenPanelId, and added active-panel event emission on state changes.

Generalized open-by-id behavior so MapView now owns wiring between toolbar UI and map state. It passes initialOpenPanelId from restoreState and listens for toolbar:activePanelChanged to update model.restoreState.openPanel.

Map URL sync now includes openPanel from model state. updateSearchParams now writes openPanel from restoreState.

src/js/views/maps/viewfinder/ViewfinderCardView.js
DarianGill: Yeah this is a good point (as usual :)) I have logic to re-open the category for a card, but not to persist the scroll position since that can be tricky based on which categories are expanded and would have required persisting the open/close state for all card categories. 

I agree the click trigger is messy and cleaning it up to implement just active styling would also reduce the DOM traversing id issues you mentioned above. I implemented this in c811fa1870ea60607d43cacb803e124b061ae227 and added regression tests to ensure the click side effects don't happen when we open an iframe and new tabs don't open when we restore a tab action. 

Now findRenderedAction() restores an now returns the actual action object, and restoreActiveAction() delegates to a new card-view restore method instead of replaying activateActionByIndex(). I got rid of activateActionByIndex() since the actual user click path is separate from the restore path and neither use it.

src/js/views/maps/viewfinder/ViewfinderView.js
DarianGill: Yeah that was to track the actual button of that action down in the DOM. Now that we're not triggering a click (as you suggested below) we don't need to do it. I changed that in the same commit: https://github.com/NCEAS/metacatui/commit/c811fa1870ea60607d43cacb803e124b061ae227

src/js/models/maps/Map.js
DarianGill: I removed the branch in 6ef5744a0c70ce7f0ec338dd89c2ad7c0c43e4f1 and added assertPlainLayerConfigs() that throws an error if a config contains backbone model instances and used it to reject preconstructed model instances during map config normalization so we fail fast if that actually happens. I also updated map-search.spec.js and map.spec.js to expect only MapAssetConfig objects and confirm we throw when model instances are provided.

src/js/models/maps/viewfinder/ViewfinderCardModel.js
DarianGill: @robyngit Hmm... I added this for the sake of compatibility with any deployed portals that already had actions in place or only updated a few of them (so we wouldn't see a strange case where some actions in certain portals were encoded and others, even potentially in the same portal, weren't) and to make testing easier for me across my demo portals. If the only place that there are actions are deployed right now is PDG, I may have been solving for a problem that doesn't exist. I thought hashing alleviated the user-updated seed value issues, but I get where you're coming from, and it is kinda complicated so I could rip it out. Out of curiosity, why didn't you seed/hash ids for layers instead of requiring an ID? Was it to encourage more complete configurations from portal managers?

src/js/views/maps/viewfinder/ViewfinderView.js
```diff
@@ -255,6 +346,8 @@ define([
         categories?.each((category) =>
           this.renderViewfinderCardsView(category),
         );
+
+        this.applyActiveActionFromUrl();
```
robyngit: When a category loads its cards from a URL, like the LEO Network, `render()` tries to restore `a=` while the collection is still empty. The fetch callback renders the cards, but does not retry restoration, so a link loads the remote cards without restoring the action.

Could we call `applyActiveActionFromUrl()` after the fetched card view has rendered?

You can reproduce the problem like so:

1. Set config to `dev.nceas`:
```js
const devnceas = {
  baseUrl: "https://dev.nceas.ucsb.edu",
  d1CNBaseUrl: "https://cn-stage.test.dataone.org/",
  metacatContext: "knb",
};
```
2. Go to portal http://localhost:3000/portals/robyn-pdg-test-2, which loads zoom presets from the LEO network: 
```json
{
    "url": "https://leonetwork.org/en/lists/geojson/A54B4AEA-21F9-4162-AEB7-AFE930C0D4E4",
    "layerIds": ["ls", "bsi"],
    "featureLayerId": "ls"
}
```
3. Click on View Layers one of the cards under latest stories, see the url update to include the active card `a`
4. Refresh the page.
5. See the card loads but the action is not selected

src/js/models/maps/viewfinder/ViewfinderCardModel.js
robyngit: The layer IDs weren't generated from a hash because hashing doesn't make a stable ID.  Changing the card description, action label or URL, layer labels, etc. changes the hash and breaks links that have already been shared or bookmarked. I think we should use the same rule here and only write `a=` for actions with an explicit ID, and leave unconfigured actions out of the URL. I'm open to more discussion though!

When the portal editor is done and supports actions, then they will automatically be assigned a stable ID via the editor.

Tried to find the discussion where we finalized the decision about only using an explicit ID for layers in links, but it's hidden in Slack's (paywalled) history... One reason why having these discussions in Github can actually be quite useful! Here is [the PR](https://github.com/NCEAS/metacatui/pull/2437) where it was implemented and here is where the decision happens [in the code](https://github.com/NCEAS/metacatui/blob/6ef5744a0c70ce7f0ec338dd89c2ad7c0c43e4f1/src/js/collections/maps/viewfinder/ViewfinderCards.js#L170-L174).

robyngit (commented):
Thanks for working through all of the earlier comments! The URL state is much easier to follow now. Just left 2 small items that I think still need to be addressed, otherwise, LGTM!

src/js/views/maps/viewfinder/ViewfinderView.js
DarianGill: Oh my, great catch! I keep missing this path--I'm glad to have [robyn-pdg-test-2](http://localhost:3000/portals/robyn-pdg-test-2) to point to for quick url fetch testing in future. I believe I fixed this in d07a4354ccc157fd45a28d5588bddc46e7a63832 by cleaning up renderViewfinderCardsView to call applyActiveActionFromUrl() at the very end after cards are populated from a category config or fetch. I also removed the top-level render() call to that function since it already calls to renderViewfinderCardsView anyway.

DarianGill:
@robyngit Re IDs:

> The layer IDs weren't generated from a hash because hashing doesn't make a stable ID. Changing the card description, action label or URL, layer labels, etc. changes the hash and breaks links that have already been shared or bookmarked. I think we should use the same rule here and only write a= for actions with an explicit ID, and leave unconfigured actions out of the URL. I'm open to more discussion though!

That's a totally valid concern--I thought you were worried about uniqueness not stability. How does that work for cards that are fetched from the leonetwork (or another URL)? Should I just pass through the id from their config? I can implement that and the change to use high res images here too, but we'd still have the challenge of making all leo-configured/url fetched cards toggle the same layers with their actions.

Eg:
```
{
      "label":"Permafrost Observations",
      "expanded":false,
      "icon":"bullhorn",
      "viewfinderCards": {
        "url":"https://leonetwork.org/en/lists/geojson/A54B4AEA-21F9-4162-AEB7-AFE930C0D4E4",
        "layerIds": [
          "ls",
          "ahri",
          "habitat-buildings",
          "habitat-roads"
        ]
      }
   },
```

Would fetch features like:
   ```
 {
      "type": "Feature",
      "geometry": {
        "type": "Point",
        "coordinates": [-162.96528, 67.57111]
      },
      "properties": {
        "id": "DF03993D-BF84-4A36-9CE3-99FECAB15A6A",
        "date": "2025-06-16T00:00:00",
        "localized_date": "Jun 16, 2025",
        "location_description": "Noatak, Alaska, United States",
        "thumbnail_url": "/en/attachments/thumbnail/533DF1E6-2F16-4E71-AEFA-E883D9DFD7A1",
        "observation": {
          "title": "Erosion results in loss of revetment wall",
          "summary": "High water on the Noatak River is accelerating erosion and causing the destruction of a decades-old cement pillow revetment wall in Noatak.",
          "source": null,
          "category_names": [
            "ENERGY",
            "EROSION",
            "FOODSECURITY",
            "INFRASTRUCTURE",
            "LAND",
            "PERMAFROST",
            "TRANSPORTATION",
            "WATER",
            "WATER_MANAGEMENT",
            "WATERSECURITY"
          ]
        }
      }
    },
```
Which would be parsed into cards like:
```
{
            "title":"Erosion results in loss of revetment wall",
            "description":"<b>2025 Jun 16:</b> High water on the Noatak River is accelerating erosion and causing the destruction of a decades-old cement pillow revetment wall in Noatak.",
            "image":"https://leonetwork.org/en/attachments/resized/533DF1E6-2F16-4E71-AEFA-E883D9DFD7A1", // <- note resized
            "buttons":[
              {
                "id":"DF03993D-BF84-4A36-9CE3-99FECAB15A6A", // <-- straight from feature ID
                "type":"map",
                "latitude":67.571115,
                "longitude":-162.96588,
                "height":1000,
                "label":"View Layers",
                "icon":"eye-open",
                "ordinality":"secondary",
                "layerIds":[  // <-- these would be pulled from the url category config and be same across all cards
                  "ls",
                  "ahri",
                  "habitat-buildings",
                  "habitat-roads",
                  "iwp"
                ]
              }
            ]
         }
         ```

robyngit:
> How does that work for cards that are fetched from the leonetwork (or another URL)? Should I just pass through the id from their config? 

Yes, I think that is a good plan since the LEO data always has an ID. Should we support other external sources for viewfinder cards in the future, those inputs will either need to adhere to our card format or we will need a custom parsing process for them. Either way, it seems feasible that we can ensure the cards always have an explicit ID. What do you think?

There's totally a trade-off between: 1. trying to support linking to any card/action even when they don't have an ID configured; and 2. creating stable links. Stipulating that all cards/actions need IDs is the simplest path and is consistent with how we've been handling linking to active layers, but a little less flexible.

DarianGill:
@robyngit Ok I got rid of the fallback button action id generation in dc17936d16633195d45cd942705dc39da14f4e10 and updated the leo network parsing in d35cfbfe48f4f6a1794e35c20035caed4f35d26c to assign actions ids and attempt to use higher res resized images for the hero. Please take another look and confirm itʻs all functional for you. I also changed the box shadows for active viewfinder cards slightly, ran prettier formatting, and updated the docs to include since 0.0.0 tags and more accurate descriptions for all new functions. :)

robyngit (approved):

PR #2869, opened 2026-08-11
Feature 2841 encode embedded app state in parent portal url

### LINKED ISSUE #2841: Forward configured portal URL parameters to embedded visualizations
## Context and motivation

Once #2635 is implemented and we can reliably navigate back to a specific visualization in a portal map, we should also allow the portal URL to include application-specific parameters that are forwarded to the embedded visualization. This way, a user may also return to the exact state of the visualization.

The initial use case is the ADC-hosted Arctic Lake Drainage Explorer, which is built with Streamlit. A shared portal URL should be able to identify the visualization and include application-specific parameters. When MetacatUI opens the iframe, those application parameters should be appended to the visualization URL. And when a visualization's state updates, those parameters should be passed back to the parent, namespaced by button action id, and encoded in the portal url.

For example, this portal URL:

`https://arcticdata.io/catalog/portals/permafrost?lt=65.16&ln=-149.58&a=wt&wt-lake_id=b7g1zd63mmt8`

could open an iframe in the portal with this URL:

`https://lostlakes.arcticdata.io/lake_id=b7g1zd63mmt8`

The visualization in the iframe can then read those values from its own query string based on its own parsing logic and restore state from the portal's url.

## Proposed generalizable configuration

To avoid making every new embedded app require a new deployment of MCUI to restore its state, we can:
1. Require an `id` for actions that sync url state that is stable over time and used to restore the iframe and namespace url encoding
2. Extend the url field in the button action config that describes the iframe to accept a standardized [rfc uri template](https://www.rfc-editor.org/info/rfc6570/) that we parse with a [library](https://github.com/lagoshny/uri-templates-es) instead of adding custom solution with path, query, and hash fields

So the lostlakes iframe ViewfinderCardAction config would look like:
```

{
  "type": "iframe",
  "id": "wt",
  "label": "Explore lake drainage",
  "url": "https://lostlakes.arcticdata.io/{+path}{?selected_lake,lat,lon,zoom,drained,month,hide_stable,show_share,theme}{#section_id}",
  "ordinality":"primary",
  "icon":"play-circle"
}
```
and we could also support /-separated paths like:

```
{
...
`"url": "https://arcticdata.io/{/selected_lake/lat/lon}{?zoom,drained,month}{#section_id}"`?
}
```

The https://github.com/lagoshny/uri-templates-es exposes a public varNames array on their [type](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/uri-templates/index.d.ts) (though it's not referenced in documentation) which creates an array of all the variables in the template we could use to whitelist certain values passed back from the iframe and reject others.

**Portal url namespace contract**

- Persisted iframe/visualization keys are namespaced by action id.
- Format: actionId-key
- Example: wt-lat, wt-lon, wt-zoom
- Existing portal-level keys like lt, ln, a remain unchanged.
- a continues to identify the active action id for restore.

## Embedded app requirements

This is the minimum an embedded visualization app must do to interoperate with MCUI deep-link state sync.

1. Parse initial state from its own URL query on load.
2. Treat those query keys as source of truth for first render.
3. On every relevant state change, update its url and send it in a postMessage to parent using the format below:

```
window.parent.postMessage(
  {
    type: "mcui:state",
    version: 1,
    url: "https://lostlakes.arcticdata.io?selected_lake=b7g1zd63mmt8lat=64.123&lon=-148.456&zoom=8&drained=true&month=7&hide_stable=false&theme=light#time-series-header"
  },
  "https://arcticdata.io"
);
```

[uri-templates-es](https://github.com/lagoshny/uri-templates-es) supports de-substitution which allows us to construct a key value object from the url in the post message, validate the keys against the accepted variables in the uri template, and then encode that state into the portal url with keys namespaced for the iframe, so we can accept postMessages like this instead of metacatui enforcing a specific params object format.

### Portal url to visualization iframe initialization (no data)

The first time we open the visualization iframe we point it to the base of the url configured in the button action, using a URL object to strip the uri query templates. 

### Visualization iFrame url to portal url synchronization

To pass information that needs to be url-encoded back from the embedded iframe to the portal we can add a PostMessage listener to MCUI. 

The messages received should have (see example above):

- `type` of "mcui:state"
- `version` of 1
- `url` a flat object of values for the keys specified in the action button config's params

Strictly speaking, we could just expect a `url` and drop the `type` and `version` requirements, but:

- `type` would allow us to skip any unrelated messages from apps/libraries (failing fast to prevent accidental parsing of foreign payloads) and extend with other messages that sync state from the portal to the iframe without reloading the page in future
- `version` would help us decide how to parse incoming post messages in the event that we change how params are handled in future. It refers only to message protocol shape and semantics and is separate from the URL parser schema version in SearchParams. A version change could look like this:

     > If we assume today (v1) iframe sends:
     > - a url that we de-substitute to get values
     > 
     > And later we want (v2):
     > - a params object with typed values
     > - full snapshot semantics
     > - omitted key means clear
     > 
     > We could increment version to 2, and the parser could handle both of these message versions simultaneously without sacrificing support for the legacy v1 type. It could also assume any message without a version is of version 1.

The MCUI post message listener would:

1. Only listen while the visualization panel is open 
      Attach on open, detach on close and onClose.
3. Check the message came from the active iframe window
      Require event.source === iframe.contentWindow
5. Strictly match origin:      
      allowedOrigin = new URL(action.url).origin      
      Require event.origin === allowedOrigin.
4. Validate the data received:
     type must be in a small whitelist, ideally only "mcui:state" for now.
     use version to see how we parse the params
5. Enforce the parameter allow-list
      Only accept values for keys specified in the configured url template.
      Drop everything else silently.
6. Parse the update object from the url in the post message
     Provided key updates value.
     Omitted allowed key is cleared from parent URL
7. Prefix the update object keys with the action id of the iframe button action 
8. Write the updated values into the portal url with history.replaceState
9. Debounce writes to reduce churn under rapid updates (eg: ideally we only update mcui url encoding when map stops moving, and only receive a post message when the embedded app's interaction has completed)
10. When the visualization is closed, remove the iframe's encoded params

Here's a code snippet for how we could use uri-template-es to make a whitelist and de-substitute an object from a uri:

```
import { UriTemplate } from 'uri-templates-es'; // 1.3.2

// 1. Initialize with the config template
const template = new UriTemplate("https://lostlakes.arcticdata.io/{+path}{?selected_lake,lat,lon,zoom,drained,month,hide_stable,theme}{#section_id}");

// 2. Officially Documented Whitelist Property
const whitelist = template.varNames;
// Outputs: ['selected_lake', 'lat', 'lon', 'zoom', 'drained', 'month', 'hide_stable', 'theme', 'section_id']

// 3. Receive a URL from the iframe postMessage
const incomingUrlFromIframe = https://lostlakes.arcticdata.io?selected_lake=b7g1zd63mmt8&lat=64.123&lon=-148.456&zoom=8&drained=true&month=7&hide_stable=false&theme=light#time-series-header;

// 4. Officially Documented Reverse-Parsing Method
const extractedState = template.fromUri(incomingUrlFromIframe); // 1.3.2

console.log(extractedState);
/*
Outputs a clean object ready for your parent site logic:
{
  selected_lake: "b7g1zd63mmt8",
  lat: 64.123,
  lon: -148.456,
  zoom: 8,
  drained: true,
  month: 7,
  hide_stable: false,
  theme: light,
  section_id: "time-series-header"
}
*/
```

### Portal url to visualization iframe restoration (with data)

When the portal is loaded and has a url with iframe data like `https://arcticdata.io/catalog/portals/permafrost?lt=65.16&ln=-149.58&a=wt&wt-lake_id=b7g1zd63mmt8` MetaCatUI would:

1. Parse any url params prefixed with the id in `a`
2. Populate the `url` template from the ViewfinderCardAction config with their parsed values
3. Preserve any already configured parameters on the iFrame's base url like `embed=true`
4. Perform the existing trusted-content validation against the final constructed URL before loading it
5. Not allow query parameters to modify the iframe URL's origin, protocol, path, or trusted-content permissions
6. Reopen the relevant iframe given by the `a` id and point it to the populated url template

For info on URL query params in streamlit, see https://discuss.streamlit.io/t/it-works-custom-copy-to-clipboard-buttons/53644 and https://discuss.streamlit.io/t/embeding-streamlit-into-iframe-with-parameters/14895/2

DarianGill:
## What
This PR closes #2841 by adding support for encoding embedded application state in the parent portal url, so shared or reloaded map links from the portal preserve that application's experience. It also renames the callbacks used by viewfinder cards to be more event driven and explicit.

## Why
The existing url encoding could open a visualization from a viewfinder card, but it did not consistently preserve the iframe-specific query context when the page was reloaded or shared. This made the map experience feel incomplete for embedded visualizations and made it harder to restore the same context from a link.

## How

The main steps in this were:

- Importing the [uri-template-es](https://github.com/lagoshny/uri-templates-es) library for parsing, substituting, and "desubstituting" uri templates
- Writing a light wrapper around the library to enforce allow-list filtering, base-URL handling, empty-value behavior, and defensive URL parsing in one place (reducing surface area if we ever need to change libraries).
- Wiring those functions into the search params and url encoding and restoration flows
- Expanding the viewfinderCardAction type to allow for:
  1. RFC6570 uri templates in the iframe `url` property that defines the accepted parameter contract between the parent and embedded app 
  2. an object property `initialQueryParams` that allows the portal manager to pass initial state to the application.

Eg:
```
{
  "title":"Near-Real Time Lake Drainage",
  "description":"Explore where lakes have drained abruptly across the Arctic in recent weeks and the last few years, most likely due to ice-rich permafrost thaw.",
  "image":"https://www.woodwellclimate.org/wp-content/uploads/2023/03/KotzebueLakeDrainage_SusanTessier.jpeg",
  "buttons":[
     {
        "id":"iframe-lost-lakes",
        "type":"iframe",
        "url": "https://lostlakes.arcticdata.io/{?selected_lake,lat,lon,zoom,drained,month,hide_stable}{#section_id}",
        "initialQueryParams":{"theme":"light", "show_share":"false"},
        "ordinality":"primary",
        "label":"Explore the Change",
        "icon":"play-circle"
     }
  ]
}
```

I added `intialQueryParams` because trying to pass those values via the `url` property required mixing the uri template with literal hard coded query params like ?theme=light, and that breaks the fromUri() destructuring that underpins the sync from the message posted from the embedded app back into the parent portal url. If we set url to something like `"https://lostlakes.arcticdata.io/?theme=light&show_share=false{?selected_lake,lat,lon,zoom,drained,month,hide_stable}{#section_id}"` the template library would expect every url posted back to the parent to also contain a `?theme=light&show_share=false` literal in the url. Since [lostlakes only ever accepts these params once and never echoes them back to the parent](https://github.com/PermafrostDiscoveryGateway/water-timeseries-v2/pull/205#issue-4904144936), the simplest way I thought of to pass them was through an additional property that's expanded into the uri template before the initial load.

Also note that the scroll position or hash id section (Time Series or Recent Imagery) isn't encoded in the url. I think this is because the hash in the lost lakes app is updated via a different event than the query params state update that's posted to the parent. I didn't stress over it because the stand alone app also doesn't seem to restore it. Visiting https://lostlakes.arcticdata.io/?selected_lake=b7g4pgc0vptj#recent-imagery for example still takes you to https://lostlakes.arcticdata.io/?selected_lake=b7g4pgc0vptj&lat=66.46608&lon=-164.18908&zoom=12#time-series-header after loading.

### Production file edits (Brought to you by MAI-Code-1-Flash)
- `package.json` and `package-lock.json `add the URI template dependency needed to resolve and expand iframe URL templates in the browser.
- `uri-templates-es.js` adds an AMD-compatible wrapper around the URI template library so MetacatUI can use it in its existing RequireJS setup.
- `app.js` registers the new URI template module path for RequireJS.
- `UriTemplateUtilities.js` adds helpers for parsing template variables, expanding templates, appending query parameters, and extracting values from visualization URLs.
- `SearchParams.js` houses the shared URL restore-state helpers for schema versioning, active action tracking, open-panel tracking, enabled-layer state, and action-specific URL synchronization.
- `Map.js` now initializes activeVisualizationAction and activeVisualizationActionId.
- `ViewfinderCardModel.js` now documents the uri template option in iframe url properties and the initialQueryParam option.
- `ViewfinderModel.js` now resolves iframe action URLs, applies initial query params, and tracks the active visualization action and URL on the map model.
- `MapView.js` wires the visualization panel to the map model and restores the panel state from the model’s active visualization attributes.
- `VisualizationPanelView.js` now handles action-aware iframe opening, validates postMessage origins, and forwards visualization state events back to the parent application.
- `ViewfinderCardView.js` now uses the renamed viewfinder card events.
- `ViewfinderCardsListView.js` now uses the renamed viewfinder card events.
- `ViewfinderView.js` restores active actions from URL state, opens the appropriate category panel, and syncs visualization state back into the parent URL using the renamed events.
- `ViewfinderCards.js` now has examples for viewfinderCard different configurations.

### Test file edits
- `SearchParams.spec.js` expands coverage for action URL resolution, namespaced URL sync, schema-based restore state, and related URL handling.
- `VisualizationPanelView.spec.js` validates the updated postMessage protocol and iframe action behavior.
-` ViewfinderCardView.spec.js` uses the renamed viewfinder card events.
- `ViewfinderCardsListView.spec.js` uses the renamed viewfinder card events.
- `ViewfinderView.spec.js` adds coverage for restoring active actions from URL state and syncing visualization state back to the URL.

## Testing
Running this branch locally and visiting `http://localhost:3000/portals/pdg_embed_vis` allows you to test the url encoding capabilities. Clicking _Explore the change_, selecting _Hide stable lakes_, and selecting lake `b7fckhphqvjs` updates the portal url to `http://localhost:3000/portals/pdg_embed_vis?lt=65.15050856557477&ln=-149.55359355073406&ht=3762014.1254774216&hd=360&p=-89.99993631512427&r=0&el=iwp&sv=1&a=iframe-lost-lakes&op=viewfinder&iframe-lost-lakes-selected_lake=b7fckhphqvjs&iframe-lost-lakes-lat=66.33621&iframe-lost-lakes-lon=-164.69712&iframe-lost-lakes-zoom=12&iframe-lost-lakes-hide_stable=1` and refreshing the page returns you to the same view in the lost lakes application.
- Ran: `npm run test`
- Result: 1415 tests passed, 0 failed

## Documentation
- The new URL restore-state contract is documented directly in the relevant code paths and JSDoc comments in SearchParams.js, ViewfinderCardModel.js, and VisualizationPanelView.js.
- The new action-level configuration for iframe URL templates and initial query params is also documented in the viewfinder card model docs and the related utility helpers.


src/js/views/maps/VisualizationPanelView.js
```diff
@@ -83,14 +84,83 @@ define(["backbone", "common/TrustedContentUtilities"], (
         this.handleEscapeKey = (e) => {
           if (e.key === "Escape") this.close();
         };
+        this.handleMessage = (event) => {
+          const iframe = this.el.querySelector(`.${CLASS_NAMES.iframe}`);
+          if (!iframe || !iframe.contentWindow) return;
+
```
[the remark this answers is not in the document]
DarianGill: Implemented for active iframe source and windows nested in iframe as well in 4aedb456df57a773958d54e48d033fc41037fbc8

src/js/common/UriTemplateUtilities.js
```diff
@@ -0,0 +1,153 @@
...
+   * @returns {object} Parsed values keyed by allowed template variable names.
+   * @since 0.0.0
+   */
+  const extractValuesFromUrl = (template, urlString) => {
+    const instance = createTemplate(template);
+    const url = toUrl(urlString);
+    if (!url) return {};
+
+    const parsed = instance.fromUri(url.toString());
+    if (!parsed || typeof parsed !== "object") return {};
```
robyngit: So if there is a parsing mismatch or error, we clear all of the current params from the URL. Is that what we want, or should we keep the URL as-is with whatever the previous params were?

src/js/common/SearchParams.js
```diff
@@ -275,6 +313,169 @@ define([], () => {
...
+  /**
+   * Build the iframe URL for an action by expanding a URI template with any
+   * namespaced portal URL state for that action.
+   * @param {object} action Viewfinder action object.
+   * @param {string} action.id Stable action identifier.
+   * @param {string} action.url RFC6570 URL template.
+   * @returns {string|null} The resolved iframe URL or null when no URL exists.
+   * @since 0.0.0
+   */
+  const resolveActionUrl = (action = {}) => {
```
robyngit: We should get the `mapModel.get("showShareUrl")` value here and if it's false, only resolve the template base and configured initial parameters, not namespaced browser state.

robyngit (approved):
Excellent! Left comments on two changes to consider, the main one being making sure we don't restore everything from the URL when `showShareUrl` is false.

src/js/common/UriTemplateUtilities.js
DarianGill: Ah interesting, I was thinking along the lines of if we get an incomplete state message we shouldn't leave "stale" params behind, but I think you're right, those stale params are probably better than nothing and we're probably better off not clobbering them when the iframe url can't be parsed. They're not really stale unless we get something valid anyway. I updated the behavior to preserve the last valid state on parse errors and added a test to exercise that path in 90a83f3c9a5bf6be50e5cebfdc87bf4731558354.

src/js/common/SearchParams.js
DarianGill: Oop yup, good idea! I added that in ec7267f98674ba8616b8715a18377049808e69f8. Only difference is I opted to call mapModel.get("showShareUrl") at the invocation site in ViewfinderModel and pass that boolean as a param so the resolveActionUrl function doesn't depend on the map model. Added a test to enforce that behavior too.

PR #2885, opened 2026-08-21
Feature 2636 url encode active feature

### LINKED ISSUE #2636: Encode active feature in cesium URL
In addition to encoding the currently active layers, we'd like the url to also encode any "active" features. This way a user can share a url that, when followed, will display the feature info panel automatically for that feature. It should also be structured such that we can support encoding multiple features in future.

This will be accomplished with a new url parameter:

1. `f` = feature(s) id(s) - used to restore selected feature(s) & feature info panel

DarianGill:
## What

This PR adds URL-based persistence for selected map features so a user can reload or share a portal and return to the same feature context. The change is centered on a schema-versioned restore-state flow in SearchParams.js, plus feature ID normalization for Cesium vector and 3D tiles layers so matches can be restored reliably across page loads. It also updates the feature info panel styling so the height adjusts smoothly across selection instead of closing and re-opening every time. Closes #2636.

## Why

The URL state already tracked camera position, enabled layers, and viewfinder panel state, but it did not include selected feature IDs. That made feature selection ephemeral and made it difficult to preserve the user’s place when a map was reloaded or shared. The problem was especially noticeable for Cesium data because generated feature IDs can be session-scoped or not stable across loads, so restoring selection by raw ID was unreliable.

### Production file edits
- `ActiveFeatureRestoreController` new module for handling async feature restore session lifecycle, feature lookup/merge logic, deferred layer waiting, and URL-sync restore-state coordination
- `SearchParams.js` adds the restore-state contract for active features encoded as json with their layer ids, and preserves unrelated URL params while writing/reading this state.
- `Map.js` still owns map state and public entrypoints like applyFeatureRestoreState() and clearFeatureRestoreSession().
- `featureIdHelpers.js` introduces shared helper logic to derive a stable, property-based feature identifier from common fields such as id, identifier, uuid, name, title, and label.
- `Cesium3DTileset.js` switches to property-based feature matching and supports deferred lookup for tiles that become visible after the layer loads. (_Did not fix all the eslint errors to minimize diff_)
- `CesiumVectorData.js` uses the same stable ID approach when looking up entities and restores selection by matching on feature properties instead of relying on Cesium-generated IDs. (_Did not fix all the eslint errors to minimize diff_)
- `FeatureInfoView.js` is updated so feature info continues to render properly when a feature is restored via the normalized ID flow.

### Test file edits

- `SearchParams.spec.js` adds coverage for restore-state parsing, URL round-tripping, and active feature ID handling.
- `Map.spec.js` covers feature selection sync to URL and feature restore behavior when layers are not yet ready.

## Testing

- `npm run test` results in 1447 passes and no failures.
- Manual testing of urls with pdg embed vis portal show reliable feature re-selection for roads, communities, local stories, virtual permafrost tours, and buildings. 

Verified with targeted branch checks:
- Ran: `npm test -- --runInBand --testPathPattern='(SearchParams|VisualizationPanelView)\.spec\.js'`
  - Result: exit code 0
- Ran: `npm test -- --runInBand --testPathPattern='models/maps/Map\.spec\.js'`
  - Result: exit code 0

## Documentation

- All new functions have since 0.0.0
- The restored state contract is documented inline in SearchParams.js.
- The stable feature-ID behavior is documented in the helper and map asset implementation files:
  - featureIdHelpers.js
  - Cesium3DTileset.js
  - CesiumVectorData.js

### DIFF
--- a/src/js/common/SearchParams.js
+++ b/src/js/common/SearchParams.js
@@ -432,7 +432,7 @@ define(["common/UriTemplateUtilities"], (UriTemplateUtilities) => {
    * @param {object} action Viewfinder action object.
    * @param {string} action.id Stable action identifier.
    * @param {string} action.url RFC6570 URL template.
-   * @param {boolean} [showShareUrl=true] Whether to read namespaced browser
+   * @param {boolean} [showShareUrl] Whether to read namespaced browser
    *   state when resolving the template.
    * @returns {string|null} The resolved iframe URL or null when no URL exists.
    * @since 0.0.0
--- a/src/js/common/SearchParams.js
+++ b/src/js/common/SearchParams.js
@@ -178,7 +178,11 @@ define(["common/UriTemplateUtilities"], (UriTemplateUtilities) => {
         : 0;
 
     // Phase 1 fields imply schema 1 writing.
-    if (normalized.activeActionId || normalized.openPanel || normalized.activeFeatureIds.length) {
+    if (
+      normalized.activeActionId ||
+      normalized.openPanel ||
+      normalized.activeFeatureIds.length
+    ) {
       normalized.schemaVersion = Math.max(1, normalized.schemaVersion);
     }
 
--- a/src/js/common/SearchParams.js
+++ b/src/js/common/SearchParams.js
@@ -97,10 +97,23 @@ define(["common/UriTemplateUtilities"], (UriTemplateUtilities) => {
    * @returns {string|null} A normalized ID string, or null if invalid.
    * @since 0.0.0
    */
-  const normalizeId = (value) => {
+  function normalizeId(value) {
     if (typeof value !== "string") return null;
     const trimmed = value.trim();
     return trimmed.length ? trimmed : null;
+  }
+
+  /**
+   * Normalize a list of IDs from array input.
+   * @param {unknown} value Candidate IDs from URL or model state.
+   * @returns {string[]} Normalized non-empty IDs.
+   * @since 0.0.0
+   */
+  const normalizeIdList = (value) => {
+    if (!Array.isArray(value)) return [];
+    return value
+      .map((item) => normalizeId(item))
+      .filter((item) => item != null);
   };
 
   /**
@@ -164,11 +177,7 @@ define(["common/UriTemplateUtilities"], (UriTemplateUtilities) => {
       enabledLayerStateProvided: Boolean(state.enabledLayerStateProvided),
       activeActionId: normalizeId(state.activeActionId),
       openPanel: normalizeId(state.openPanel),
-      activeFeatureIds: parseCommaSeparated(
-        Array.isArray(state.activeFeatureIds)
-          ? state.activeFeatureIds.join(",")
-          : state.activeFeatureIds,
-      ),
+      activeFeatureIds: normalizeIdList(state.activeFeatureIds),
     };
 
     const requestedSchema = Number(state.schemaVersion);
@@ -229,8 +238,8 @@ define(["common/UriTemplateUtilities"], (UriTemplateUtilities) => {
     if (schemaVersion >= 1) {
       base.activeActionId = normalizeId(url.searchParams.get(ACTIVE_ACTION_ID));
       base.openPanel = normalizeId(url.searchParams.get(OPEN_PANEL_ID));
-      base.activeFeatureIds = parseCommaSeparated(
-        url.searchParams.get(ACTIVE_FEATURES_ID),
+      base.activeFeatureIds = normalizeIdList(
+        url.searchParams.getAll(ACTIVE_FEATURES_ID),
       );
     }
 
@@ -283,10 +292,9 @@ define(["common/UriTemplateUtilities"], (UriTemplateUtilities) => {
         url.searchParams.set(OPEN_PANEL_ID, normalized.openPanel);
       }
       if (normalized.activeFeatureIds.length) {
-        url.searchParams.set(
-          ACTIVE_FEATURES_ID,
-          normalized.activeFeatureIds.join(","),
-        );
+        normalized.activeFeatureIds.forEach((featureId) => {
+          url.searchParams.append(ACTIVE_FEATURES_ID, featureId);
+        });
       }
     }
 
--- a/src/js/common/SearchParams.js
+++ b/src/js/common/SearchParams.js
@@ -16,6 +16,7 @@ define(["common/UriTemplateUtilities"], (UriTemplateUtilities) => {
   const SCHEMA_VERSION_ID = "sv";
   const ACTIVE_ACTION_ID = "a";
   const OPEN_PANEL_ID = "op";
+  const ACTIVE_FEATURES_ID = "f";
 
   /** The search parameter ID for enabled layers in the save to URL feature. */
   const ENABLED_LAYERS_ID = "el";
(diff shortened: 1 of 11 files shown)


src/js/common/SearchParams.js
```diff
@@ -161,6 +164,11 @@ define(["common/UriTemplateUtilities"], (UriTemplateUtilities) => {
       enabledLayerStateProvided: Boolean(state.enabledLayerStateProvided),
       activeActionId: normalizeId(state.activeActionId),
       openPanel: normalizeId(state.openPanel),
+      activeFeatureIds: parseCommaSeparated(
+        Array.isArray(state.activeFeatureIds)
+          ? state.activeFeatureIds.join(",")
+          : state.activeFeatureIds,
```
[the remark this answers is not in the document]
DarianGill: Resolved in cdb61c2960c5928c8461b5e2cf9c4ed618d94212

src/js/models/maps/Map.js
```diff
@@ -709,6 +825,137 @@ define([
...
+        if (alreadySelected) {
+          this.clearFeatureRestoreSession();
+          return;
+        }
+
+        const featureAttrs = this.findFeatureAttributesByIds(activeFeatureIds);
+        if (featureAttrs.length) {
+          this.selectFeatures(featureAttrs);
+          this.clearFeatureRestoreSession();
+          return;
```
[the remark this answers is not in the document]
DarianGill: Addressed in 70577016c83dbaf308d32d73e9c0c058192e0a8e

src/js/models/maps/featureIdHelpers.js
```diff
@@ -0,0 +1,69 @@
...
+    const normalizedId = id != null ? String(id).trim() : "";
+    if (!normalizedId) {
+      return false;
+    }
+
+    const lower = toLowerCaseProps(props);
+    return FEATURE_ID_KEYS.some((key) => {
+      const val = lower[key];
+      return val != null && String(val).trim() === normalizedId;
+    });
```
robyngit:
```suggestion
    return getIdFromProperties(props) === id;
```

When creating a URL, we choose one preferred identifier key for a feature. When reading that URL, we accept several different properties as matches. That can make the map open a different feature from the one that was shared.

For example, a layer contains these features in this order:

```
featureA = { id: "A", name: "B" };
featureB = { id: "B" };
```

Selecting `featureB` in the UI adds `?f=B` to the URL.

On reload, the search reaches `featureA` first. Its name is "B", so the reader says it matches. The panel opens for `featureA`, even though its canonical ID is "A".

This is not likely to happen, since it requires one feature's name, title, label, etc. to equal another feature's canonical ID, but still better to use the same algorithm for matching when writing the URL as when reading the URL.

src/js/views/maps/FeatureInfoView.js
```diff
@@ -381,7 +396,14 @@ define([
...
 
-            let searchKeys = ["name", "title", "id", "identifier"];
+            let searchKeys = [
+              "name",
+              "title",
+              "label",
+              "uuid",
+              "id",
+              "identifier",
+            ];
```
robyngit: Can we import these from the constant in `featureIdHelpers` so there's 1 source of truth for the search keys?

src/js/views/maps/FeatureInfoView.js
```diff
@@ -250,11 +258,18 @@ define([
         // Insert the title into the title element
         this.elements.title.innerHTML = title;
 
+        // Collapse only when opening from a closed state. When switching between
+        // selected features, keep the current height and transition directly to the
+        // new content height.
+        if (collapseBeforeLoad) {
+          iFrame.style.height = 0;
+          iFrame.style.opacity = 0;
```
robyngit: nitpick: maybe a small `hideiFrame` function since we do this twice?

src/js/models/maps/ActiveFeatureRestoreController.js
```diff
@@ -0,0 +1,337 @@
...
+          const feature = layer.getFeatureById(id);
+          if (!feature) return foundAttrs;
+
+          return layer.getFeatureAttributes(feature) || foundAttrs;
+        }, null);
+
+        if (featureAttrs) result.push(featureAttrs);
+        return result;
+      }, []);
+    },
```
robyngit, line 344: The URL records which feature was selected, but not which layer it came from. If two layers use the same feature ID, the map will restore whichever layer it searches first. I think we need to serialize the selected feature id along with the layer ID it's from. We could add a more reusable/generic method to the Map model, like `findFeature(featureId, layerId)`.

I think it's probably this collision in IDs could happen, say in layers that use a simple row number for ID.

src/js/models/maps/assets/CesiumVectorData.js
```diff
@@ -401,7 +390,12 @@ define([
...
         feature = this.getEntityFromMapObject(feature);
         if (!feature) return null;
-        return feature.id;
+        // Prefer a stable property-based ID over Cesium's auto-generated entity
+        // UUID (which changes on every page load).
+        return (
+          getIdFromProperties(this.getPropertiesFromFeature(feature)) ??
+          feature.id
+        );
       },
```
robyngit, line 399: Because this falls back to `feature.id`, `featureID` can contain a Cesium-generated UUID when the feature has no recognized ID property. `getSelectedFeatureIdsForUrlState()` currently serializes every nonempty string ID, so that unstable UUID can end up in the URL and no longer identify the feature after reload.

The fallback still seems appropriate here because this method provides the feature's general in-session ID and preserves the original behaviour. Could the URL serialization path get the stable ID from the feature's properties with getIdFromProperties() and omit the feature when none exists? That would be in keeping with the "stable IDs only" rule

robyngit (changes requested):
This looks good @DarianGill! Three things to fix then it's ready to go IMO (details in comments):
1. IDs should be resolved the same way when writing and restoring.
2. IDs need to be scoped to their layers.
3. Unstable/auto-generated IDs can still make it into URLs.

src/js/models/maps/featureIdHelpers.js
DarianGill: Oop yeah good edge case catch. That does sound like the better approach. Implemented in 79456499f7aa2c086230d25defe16e9a153b2773 but kept the normalization. Added tests for the helper functions to ensure this and other behavior is enforced.

src/js/views/maps/FeatureInfoView.js
DarianGill: Yup! That was redundant. As of 745eb3f7be1813cda2003b71745893632524d98f we reuse them.

src/js/views/maps/FeatureInfoView.js
DarianGill: Done in 468fe054bc51cdc7356f32c694868f30c3ef9f3a as collapseiFrame.

src/js/models/maps/ActiveFeatureRestoreController.js
DarianGill: Yeah good point. I updated it to use a json encoded string with the feature and layer ids in 8d788159a5f7d099c1a7c369eb073c9a112ee344.

src/js/models/maps/assets/CesiumVectorData.js
DarianGill: I believe I implemented that in f3249fbd9a1d72d84e1265985f498db8e538bb32

robyngit (approved):
Excellent @DarianGill! Ready to merge, but please wait until I've got the release out :)

DarianGill:
Thank you! Will do :)

Issue #2126, opened 2023-04-12
Indicate that a layer is updating/loading in Cesium


robyngit:
Some layers in Cesium take some time to update or load (e.g. the Geohash layer at the moment). It would be helpful to show some sort of subtle indicator that the layer is loading. For example, TerriaJS shows a simple white progress bar at the very top of the map. You can see this by just panning around on https://nationalmap.gov.au. I think a similar strategy would work well for our map.

DarianGill:
Ahg dyslexia got the better of me and I've been naming my branch and tagging my commits 2162 instead of 2126 🤦

Issue #2870, opened 2026-08-12
Consolidate url state encoding orchestration


DarianGill:
This issue is for us to consider an overall architecture tidy-up of how the URL state is managed. We've slowly pulled the url functionality out of certain modules like `CesiumWidgetView`, but the functionality remains spread out across a wide surface that can be hard to reason about. Right now, `SearchParams` reads and writes the browser URL, `ViewfinderModel` resolves and opens actions, `ViewfinderView` owns the debounce and sync, and `MapView`/`VisualizationPanelView` relay the state. Same for the active visualization, which is right now split across `activeVisualizationAction`, `activeVisualizationActionId`, and `activeVisualizationUrl`, which always need to be updated at the same time. We should find a way we can consolidate the URL coordination into a single place and make clearer ownership boundaries.

 Here's one way we could do that (conceived with the help of GPT-5.3-codex)

## Plan: Single URL Coordinator + Unified Visualization State

The cleanest approach is to make one coordinator the only owner of URL state, and make active visualization a single atomic object. The following could be done in one big refactor PR:

### Steps
1. Phase 1: Establish one URL owner (blocks all other phases).
2. Add a coordinator module that is the only code allowed to read, normalize, and write browser query state.
3. Move URL schema logic out of SearchParams.js into the coordinator internals (camera, layers, open panel, active action id, namespaced action params, schema versioning).
4. Keep SearchParams.js as a temporary facade that delegates to coordinator methods to limit import churn during this PR.
5. Parse URL once on app/map init, cache normalized state in coordinator memory, and expose it to consumers so views/models do not repeatedly parse window location.
6. Phase 2: Unify active visualization state (depends on Phase 1).
7. Replace Map attributes activeVisualizationAction, activeVisualizationActionId, and activeVisualizationUrl with one activeVisualization object in Map.js.
8. Refactor write sites to set activeVisualization atomically in ViewfinderModel.js and MapView.js.
9. Refactor read sites to consume activeVisualization atomically in MapView.js, ViewfinderView.js, and VisualizationPanelView.js.
10. Add temporary read-only compatibility aliases in Map for one PR so untouched callers do not break mid-refactor.
11. Phase 3: Move URL sync orchestration into coordinator (depends on Phases 1-2).
12. Replace distributed URL writes from Map/Viewfinder/MapView with coordinator intent methods (camera changed, layers changed, panel changed, visualization opened/closed, visualization state message received).
[...]

PR #2888, opened 2026-08-29
Feature 2126 layer loading indicator

DarianGill:
## What
Closes [#2126](https://github.com/NCEAS/metacatui/issues/2126) by adding a status bar widget at the top-right of the map that shows a loading row beneath the scale bar whenever layers are loading or being rendered for the first time. The row shows a layer name (or a summary count for multiple layers) and collapses automatically once all tracked layers are ready. Layer items in the toolbar also reflect the per-layer loading state through a shared `isLoadingLayer` flag. An internal `LayerLoadingCoordinator` module centralizes all logic for computing the aggregate loading state.

Additionally, the feature-restore path is hardened so stale asynchronous waiters expire after a bounded timeout, layer visibility changes are handled correctly (interrupting only the affected feature's restore, restarting when a hidden searchable layer becomes visible again), and error/invisible layers are excluded from restore searches.

Also fixes two related UI polish issues that came up during development: the layer toolbar's loading/error icon and settings button no longer shift neighboring icons when they appear/disappear, and the in-progress draw polygon no longer gets visually buried underneath other vector layers that load in later.

## Why
Previously there was no map-wide visual feedback while layers were loading. Users could open the map and see either a blank globe or partial data with no indication that more content was incoming. The layer toolbar's loading spinner was also inconsistent—it only reflected the data-request phase (`status === "loading"`), not the additional render phase (`displayReady === false`/`tilesLoading === true`) that occurs after data arrives.

The feature-restore session also had subtle correctness issues: it could hang indefinitely if no matching feature ever loaded, would retry against invisible or errored layers, and did not restart when the set of visible searchable layers changed between calls.

Two secondary issues surfaced during manual testing: rapid layer status churn made the status bar's loading message flicker between names/counts, and the layer toolbar's status icon/settings button toggling `display` caused neighboring icons (filter icon, gear button) to jostle horizontally. Separately, because the draw layer's `CustomDataSource` is created once, early, in `DownloadPanelView.initialize()`, any other vector layer added or toggled visible afterward could end up stacked above it in Cesium's draw order, burying the user's in-progress polygon.

## How (Claude Sonnet 5)
**Production files**

- `LayerLoadingCoordinator.js` — **New module.** A pure, stateless computation module that extracts all aggregate loading-state logic out of `Map.js`: determines which layers should be tracked (`excludeFromLoadingState`), identifies layers that are still loading (`status === "loading"`, or `tilesLoading === true` when tracked, or the `displayReady === false` fallback for layer types that don't track `tilesLoading`, e.g. vector data), formats the user-facing loading message (one label, or "X and N other layer(s)"), truncates long/HTML-bearing labels without breaking markup, syncs the per-layer `isLoadingLayer` flag, and writes `isLoadingLayers`/`loadingLayersMessage` to the map model.
- `MapStatusBarView.js` — **New view.** Composes `ScaleBarView` with the new `LayerLoadingIndicatorView` and owns all reveal/collapse timing: waits 1 s before expanding the loading row (so brief loads never flash), keeps it open at least 1 s once expanded, shows a "Loading completed!" message if loading finishes early, and mirrors the scale bar's live pixel width onto a CSS variable so the box grows/shrinks by exactly that amount. Also throttles visible message-text updates via `scheduleMessageUpdate()`/`flushMessageUpdate()` to once per `MESSAGE_UPDATE_INTERVAL_MS` (400 ms), so a burst of near-simultaneous layer status changes settles into one update instead of flickering, while always resolving to the latest message once the window elapses. A fresh loading start/resume (tracked via `isCurrentlyLoading`) and the one-time "Loading completed!" message both bypass the throttle so meaningful transitions stay immediate; the new timer is cleaned up in `handleLoadingFinished()`/`onClose()`.
- `LayerLoadingIndicatorView.js` — **New view.** A purely presentational animated bar + message row with no timing logic of its own; sets the message via `innerHTML` so raw HTML labels (e.g. `<sub>`/`<sup>`) render the same as they do in the layer menu.
- `Map.js` — Imports `LayerLoadingCoordinator` and listens on each layer-group collection (`change:status`/`change:displayReady`/`change:tilesLoading`, `change:visible`, and `update`/`reset`) to recompute loading state and re-run feature restore as needed. Adds `featureRestoreTimeoutMs` (default `15000`) to defaults. Marks the default base layer with `excludeFromLoadingState: true` so it never contributes to the loading widget. Hiding a layer mid-restore now clears just that layer's restore entries/selected features instead of the whole session.
- `ActiveFeatureRestoreController.js` — Adds a bounded session timeout that calls `LayerLoadingCoordinator.updateLayerLoadingState` on expiry. Incorporates the current searchable/visible layer ids into the session key (`serializeRestoreScopeKey`, JSON-serialized to avoid comma-collision) so visibility changes invalidate the existing session. Filters out invisible layers, error-status layers, and non-observable objects from restore searches and waiter registrations. Adds `clearFeatureRestoreEntriesForLayer()` to surgically remove one layer's restore entries and re-sync the URL.
- `MapAsset.js` — Adds `displayReady` (default `null`, one-time latch) and `tilesLoading` (default `null`, continuously-updated pending-work signal) attributes. `resetStatus()` now sets `status` to `"loading"`, `displayReady` to `false`, and `tilesLoading` to `null` rather than resetting to defaults. Updates the JSDoc for `status` to include the `'loading'` value. Adds `startLoadingStateTracking()`/`stopLoadingStateTracking()` no-op hooks for subclasses to override. Adds a `renderAboveOtherLayers` flag to defaults, used to mark layers (like the in-progress draw polygon) that should always be re-raised above other vector layers as they load in.
- `Cesium3DTileset.js` — Back-references the owning `MapAsset` model on the Cesium primitive (`cesiumModel.mapAssetModel`). Implements `startLoadingStateTracking()`/`stopLoadingStateTracking()`: sets `displayReady` on the tileset's `tileVisible` event (one-time), and keeps `tilesLoading` in sync via the tileset's own `loadProgress` event (`numberOfPendingRequests`/`numberOfTilesProcessing`) for the layer's whole lifetime in the scene.
- `CesiumImagery.js` — Back-references the `MapAsset` model on the Cesium imagery layer. Implements `startLoadingStateTracking()`/`stopLoadingStateTracking()`: `isDisplayReadyInScene()` inspects the imagery cache for a `READY`/`TEXTURE_LOADED` tile to set the one-time `displayReady` latch on `postRender`. `wrapRequestImageForPendingCount()` monkey-patches the provider's `requestImage` to count only genuinely in-flight requests for `tilesLoading`, avoiding false "still loading" state from Cesium's internal cache retaining stale/orphaned LOD-fallback tiles.
- `CesiumVectorData.js` — Fixes a bug where `model.setError.bind(...)` was never invoked (leaving `status: "loading"` stuck forever on load failure); now calls `model.setError(...)` directly. Only resets `displayReady` during an active load cycle (`status === "loading"`), not during selection/highlight restyling. Also tags the Cesium `CustomDataSource` with `mapAssetModel` so `CesiumWidgetView` can look up the asset config for a given data source.
- `CesiumWidgetView.js` — In `add3DTileset`/`addImagery`, calls the asset's `startLoadingStateTracking({ scene })` and requests a render. In `remove3DTileset`/`removeImagery`, calls `stopLoadingStateTracking()` to clean up listeners. `addVectorData`/`removeVectorData` now track any data sources flagged `renderAboveOtherLayers` and re-raise them to the top of the `dataSourceCollection` every time a vector layer is added, so a layer like the draw polygon keeps rendering above layers that load in later.
- `MapWidgetContainerView.js` — `renderScaleBar()` renamed to `renderStatusBar()` and now renders `MapStatusBarView` (passing the map model) in place of a bare `ScaleBarView`; `this.scaleBar` renamed to `this.statusBar`.
- `ScaleBarView.js` — Hides the lat/lng labels with `visibility` instead of `display` so they keep occupying layout space (preventing shift on mouse enter/leave), and triggers `update:barWidth` whenever the bar's pixel width changes so `MapStatusBarView` can mirror it.
- `MapView.js` — Passes `mapModel` to `FeatureInfoView` and clears the (now-removed) loading indicator timer reference in `onClose`.
- `LayerItemView.js` — `showStatus()` now reads both `status` and `isLoadingLayer` (via new `getStatusState()`/`appendStatusElement()` helpers) to decide what to render, so the loading spinner appears whenever the map model marks the layer as loading, not only when `status === "loading"`. Adds a listener on `change:isLoadingLayer` and re-evaluates status on `change:visible`. Nulls out `statusIcon`/`badge` references in `removeStatuses()`. Also now creates a fixed-size, always-present `.list-item__status-icon` slot during `render()` (before any loading/error state is known); `showLoading()`/`showError()` just set `innerHTML` on that slot and toggle a `--visible` modifier class, and `removeStatuses()` clears it back out—the element itself never enters/leaves the DOM, so it no longer jostles the filter icon.
- `FeatureInfoView.js` — Calls `mapModel.clearFeatureRestoreSession()` when the feature info panel is closed, so a dismissed panel doesn't leave a stale restore session active.
- `DownloadPanelView.js` — Marks the ephemeral drawing polygon layer with `excludeFromLoadingState: true` (so it never triggers the loading indicator) and `renderAboveOtherLayers: true` (so it stays visible above other layers while the user draws).
- `map-view.css` — Replaces the old `.scale-bar` self-positioning box with a new `.map-status-bar` wrapper that owns position/shadow/background/z-index and an explicit, non-content-driven width (base width + live bar-width CSS variable) so it can animate smoothly with `transition: width` in every browser. Adds the `.map-status-bar__loading-row` grid-rows (`0fr`→`1fr`) expand/collapse animation, `__loading-bar` (animated sliding highlight), `__loading-message`/`__loading-text` styles. Adds tabular-nums + reserved min-widths on the lat/lng coordinate spans to prevent jitter while panning. Increases `.map-view__feature-info-container`'s reserved top margin (`max-height`) so a tall panel can't grow up far enough to cover the status bar, and switches it from `overflow: hidden` to scrollable. Gives `.list-item__status-icon` a fixed 1rem × 1rem footprint with `visibility: hidden` by default (reserving the space) and `visibility: visible` when active. `.list-item__settings` (the gear/settings button) now stays `display: flex` at all times and toggles `visibility: hidden` ↔ `visible` instead, with a permanent (transparent-when-hidden) 1px border so its box size never changes and the grid column width stays constant regardless of hover state.

**Test files**

- `MapStatusBarView.spec.js` — **New spec.** Covers initialization/rendering and the full timing state machine with `sinon.useFakeTimers()`: no expansion before the reveal delay, expansion after the delay, no flash if loading finishes first, minimum-open-duration + "completed" message, immediate collapse after the minimum duration has elapsed, and staying open/updating the message if loading resumes before collapsing.
- `Map.spec.js` — Adds a suite of tests covering: default base layer excluded from loading tracking; feature restore retried when layer loading metadata changes; restore sessions replaced when visible searchable layers change (including comma-containing ids and stale-waiter cancellation); restore state kept alive when an unrelated layer is hidden but cleared/URL-synced when the restoring layer itself is hidden; restore sessions never masquerade as layer loading state; visible loading layers set `isLoadingLayers` while error/not-yet-display-ready layers are handled correctly; per-layer `isLoadingLayer` flags stay in sync, including on dynamic add/remove; helper layers with `excludeFromLoadingState` are ignored; multi-layer loading messages are formatted correctly; labels preserve `<sub>`/`<sup>` markup and truncate long labels without breaking markup.
- `MapView.spec.js` — Updates the feature-info suite to verify `clearFeatureRestoreSession()` is called on close; removes the old in-`MapView` loading indicator timing suite (that DOM now lives in `MapStatusBarView`).
- `LayerItemView.spec.js` — Adds a `status rendering` suite verifying the loading spinner appears only when `isLoadingLayer` is true and the layer is visible.
- `MapWidgetContainerView.spec.js` — Updates scale-bar tests to assert `.map-status-bar` presence/absence instead of the old bare `.scale-bar` container.

## Testing

- 1479 tests passing, 0 failures (`PORT=3002 npm test`).
- All new behavior is covered by unit tests in `Map.spec.js`, `MapStatusBarView.spec.js`, and `LayerItemView.spec.js`.
- Manual: load a map with several tile/imagery layers; the status bar's loading row should expand ~1 s after a layer starts loading and list the loading layer name(s), then collapse (with a brief "Loading completed!" message if it finishes early) once all visible layers have rendered at least one tile/imagery frame.
- Verify that toggling a layer off while it is still loading removes it from the indicator, and toggling it back on restores it.
- Verify the drawing polygon layer and the default base layer never appear in the loading message.
- Verify panning the map (bringing new imagery/3D tiles into view) doesn't cause the indicator to get stuck open or closed.
- Verify that with several layers rapidly changing loading status in quick succession, the status bar message settles smoothly instead of flickering.
- Verify the layer toolbar's filter icon and settings/gear button no longer shift position when a layer's loading/error icon appears or disappears.
- Verify the in-progress draw polygon stays visible above other vector layers even as those layers load in or are toggled on after drawing has started.

## Documentation
No user-facing documentation changes are required. JSDoc on `MapAsset.status` is updated to document the new `'loading'` value. New public methods carry JSDoc comments with `@since 0.0.0` placeholders. `Map.js` defaults JSDoc includes the new `featureRestoreTimeoutMs` property.

### DIFF
--- a/src/js/models/maps/ActiveFeatureRestoreController.js
+++ b/src/js/models/maps/ActiveFeatureRestoreController.js
@@ -306,6 +306,11 @@ define(["backbone", "models/maps/LayerLoadingCoordinator"], (
         return;
       }
 
+      if (!allSearchableLayers.length) {
+        this.clearSession();
+        return;
+      }
+
       if (!canResolveAsynchronously) {
         this.clearSession();
         mapModel.syncSelectedFeaturesToUrl();
--- a/src/js/models/maps/ActiveFeatureRestoreController.js
+++ b/src/js/models/maps/ActiveFeatureRestoreController.js
@@ -51,6 +51,20 @@ define(["backbone"], (Backbone) => {
     this.mapModel = mapModel;
   }
 
+  /**
+   * Check whether a layer supports Backbone-style event listening.
+   * @param {object} layer Candidate layer object.
+   * @returns {boolean} True when the layer has on/off event functions.
+   * @since 0.0.0
+   */
+  function isObservableLayer(layer) {
+    return (
+      layer &&
+      typeof layer.on === "function" &&
+      typeof layer.off === "function"
+    );
+  }
+
   MapFeatureRestoreController.prototype = {
     /**
      * Get the current restore session from the owning map.
@@ -107,21 +121,33 @@ define(["backbone"], (Backbone) => {
     /**
      * Start a new feature restore session, canceling any previous one.
      * @param {string[]} activeFeatureIds The ids being restored.
+     * @param {string} [sessionKey] Optional key describing restore scope.
      * @returns {object} The active restore session.
      * @since 0.0.0
      */
-    beginSession(activeFeatureIds) {
-      const sessionKey = activeFeatureIds.join(",");
+    beginSession(activeFeatureIds, sessionKey = activeFeatureIds.join(",")) {
       if (this.getSession()?.key === sessionKey) {
         return this.getSession();
       }
 
       this.clearSession();
-      return this.setSession({
+      const session = this.setSession({
         cancelers: [],
         key: sessionKey,
         requestedIds: activeFeatureIds.slice(),
       });
+
+      // Ensure restores have a bounded lifetime so unresolved waits do not
+      // persist forever when no matching feature ever arrives.
+      const timeoutMs = this.mapModel.get("featureRestoreTimeoutMs") || 15000;
+      const timeoutId = setTimeout(() => {
+        if (!this.isActiveSession(session)) return;
+        this.clearSession();
+        this.mapModel.updateLayerLoadingState();
+      }, timeoutMs);
+      this.addWaiter(() => clearTimeout(timeoutId), session);
+
+      return session;
     },
 
     /**
@@ -154,14 +180,13 @@ define(["backbone"], (Backbone) => {
      * Search all layers for features matching the given ids and return
      * feature attribute objects ready to be passed to selectFeatures().
      * @param {string[]} ids Feature ids to search for.
+     * @param {Array<object>} [layers] Layers to search within.
      * @returns {object[]} Matching feature attribute objects.
      * @since 0.0.0
      */
-    findFeatureAttributesByIds(ids) {
-      const allLayers = this.mapModel.getAllLayers();
-
+    findFeatureAttributesByIds(ids, layers = this.mapModel.getAllLayers()) {
       return ids.reduce((result, id) => {
-        const featureAttrs = allLayers.reduce((foundAttrs, layer) => {
+        const featureAttrs = layers.reduce((foundAttrs, layer) => {
           if (
             foundAttrs ||
             typeof layer.getFeatureById !== "function" ||
@@ -211,10 +236,14 @@ define(["backbone"], (Backbone) => {
         .getAllLayers()
         .filter(
           (layer) =>
+            layer.get("visible") !== false &&
             typeof layer.getFeatureById === "function" &&
             layer.get("status") !== "error",
         );
-      const featureAttrs = this.findFeatureAttributesByIds(activeFeatureIds);
+      const featureAttrs = this.findFeatureAttributesByIds(
+        activeFeatureIds,
+        allSearchableLayers,
+      );
       const resolvedIds = new Set(selectedRequestedIds);
 
       featureAttrs.forEach((feature) => {
@@ -228,6 +257,10 @@ define(["backbone"], (Backbone) => {
         (id) => !resolvedIds.has(id),
       );
       const restoreKey = activeFeatureIds.join(",");
+      const searchableLayerKey = allSearchableLayers
+        .map((layer) => layer.get("layerId") || layer.cid || "")
+        .join(",");
+      const restoreScopeKey = `${restoreKey}|${searchableLayerKey}`;
       const canResolveAsynchronously = allSearchableLayers.some(
         (layer) =>
           layer.get("status") !== "ready" ||
@@ -239,9 +272,9 @@ define(["backbone"], (Backbone) => {
       if (
         unresolvedIds.length &&
         canResolveAsynchronously &&
-        existingSession?.key !== restoreKey
+        existingSession?.key !== restoreScopeKey
       ) {
-        restoreSession = this.beginSession(activeFeatureIds);
+        restoreSession = this.beginSession(activeFeatureIds, restoreScopeKey);
       }
 
       if (featureAttrs.length) {
@@ -261,7 +294,10 @@ define(["backbone"], (Backbone) => {
         return;
       }
 
-      if (this.getSession()?.key === restoreKey && existingSession?.key === restoreKey) {
+      if (
+        this.getSession()?.key === restoreScopeKey &&
+        existingSession?.key === restoreScopeKey
+      ) {
         return;
       }
 
@@ -277,7 +313,10 @@ define(["backbone"], (Backbone) => {
           return;
         }
 
-        const attrs = this.findFeatureAttributesByIds(activeFeatureIds);
+        const attrs = this.findFeatureAttributesByIds(
+          activeFeatureIds,
+          allSearchableLayers,
+        );
         if (attrs.length) {
           mapModel.selectFeatures(
             mergeFeatureSelections(
@@ -319,6 +358,11 @@ define(["backbone"], (Backbone) => {
 
       allSearchableLayers.forEach((layer) => {
         if (layer.get("status") !== "ready") {
+          if (!isObservableLayer(layer)) {
+            registerTileWaiters(layer);
+            return;
+          }
+
           const statusListener = () => {
             if (!this.isActiveSession(restoreSession)) {
               mapModel.stopListening(layer, "change:status", statusListener);
--- a/src/js/models/maps/ActiveFeatureRestoreController.js
+++ b/src/js/models/maps/ActiveFeatureRestoreController.js
@@ -1,6 +1,9 @@
 "use strict";
 
-define(["backbone", "models/maps/LayerLoadingCoordinator"], (Backbone, LayerLoadingCoordinator) => {
+define(["backbone", "models/maps/LayerLoadingCoordinator"], (
+  Backbone,
+  LayerLoadingCoordinator,
+) => {
   /**
    * Extract a feature id from either a Feature model or plain attrs object.
    * @param {Backbone.Model|object} feature Feature model or attrs object.
@@ -59,9 +62,7 @@ define(["backbone", "models/maps/LayerLoadingCoordinator"], (Backbone, LayerLoad
    */
   function isObservableLayer(layer) {
     return (
-      layer &&
-      typeof layer.on === "function" &&
-      typeof layer.off === "function"
+      layer && typeof layer.on === "function" && typeof layer.off === "function"
     );
   }
 
--- a/src/js/models/maps/ActiveFeatureRestoreController.js
+++ b/src/js/models/maps/ActiveFeatureRestoreController.js
@@ -1,6 +1,6 @@
 "use strict";
 
-define(["backbone"], (Backbone) => {
+define(["backbone", "models/maps/LayerLoadingCoordinator"], (Backbone, LayerLoadingCoordinator) => {
   /**
    * Extract a feature id from either a Feature model or plain attrs object.
    * @param {Backbone.Model|object} feature Feature model or attrs object.
@@ -143,7 +143,7 @@ define(["backbone"], (Backbone) => {
       const timeoutId = setTimeout(() => {
         if (!this.isActiveSession(session)) return;
         this.clearSession();
-        this.mapModel.updateLayerLoadingState();
+        LayerLoadingCoordinator.updateLayerLoadingState(this.mapModel);
       }, timeoutMs);
       this.addWaiter(() => clearTimeout(timeoutId), session);
 
--- a/src/js/models/maps/ActiveFeatureRestoreController.js
+++ b/src/js/models/maps/ActiveFeatureRestoreController.js
@@ -44,6 +44,21 @@ define(["backbone", "models/maps/LayerLoadingCoordinator"], (
   }
 
   /**
+   * Serialize a feature-restore scope without ambiguity between ids that may
+   * contain separator characters.
+   * @param {string[]} featureIds Requested feature ids.
+   * @param {string[]} layerIds Searchable layer ids.
+   * @returns {string} Stable scope key.
+   * @since 0.0.0
+   */
+  function serializeRestoreScopeKey(featureIds, layerIds) {
+    return JSON.stringify({
+      featureIds: [...new Set(featureIds)].sort(),
+      layerIds: [...new Set(layerIds)].sort(),
+    });
+  }
+
+  /**
    * Manage asynchronous feature restore state for a map model.
    * @param {object} options Controller options.
    * @param {MapModel} options.mapModel Owning map model.
@@ -257,11 +272,13 @@ define(["backbone", "models/maps/LayerLoadingCoordinator"], (
       const unresolvedIds = activeFeatureIds.filter(
         (id) => !resolvedIds.has(id),
       );
-      const restoreKey = activeFeatureIds.join(",");
-      const searchableLayerKey = allSearchableLayers
+      const searchableLayerIds = allSearchableLayers
         .map((layer) => layer.get("layerId") || layer.cid || "")
-        .join(",");
-      const restoreScopeKey = `${restoreKey}|${searchableLayerKey}`;
+        .filter((layerId) => typeof layerId === "string" && layerId.length);
+      const restoreScopeKey = serializeRestoreScopeKey(
+        activeFeatureIds,
+        searchableLayerIds,
+      );
       const canResolveAsynchronously = allSearchableLayers.some(
         (layer) =>
           layer.get("status") !== "ready" ||
--- a/src/js/models/maps/ActiveFeatureRestoreController.js
+++ b/src/js/models/maps/ActiveFeatureRestoreController.js
@@ -162,7 +162,11 @@ define(["backbone"], (Backbone) => {
 
       return ids.reduce((result, id) => {
         const featureAttrs = allLayers.reduce((foundAttrs, layer) => {
-          if (foundAttrs || typeof layer.getFeatureById !== "function") {
+          if (
+            foundAttrs ||
+            typeof layer.getFeatureById !== "function" ||
+            layer.get("status") === "error"
+          ) {
             return foundAttrs;
           }
 
@@ -203,9 +207,13 @@ define(["backbone"], (Backbone) => {
       const selectedRequestedIds = (selectedFeatures?.models || [])
         .map((feature) => getFeatureId(feature))
         .filter((id) => activeFeatureIds.includes(id));
-      const allSearchableLayers = mapModel.getAllLayers().filter(
-        (layer) => typeof layer.getFeatureById === "function",
-      );
+      const allSearchableLayers = mapModel
+        .getAllLayers()
+        .filter(
+          (layer) =>
+            typeof layer.getFeatureById === "function" &&
+            layer.get("status") !== "error",
+        );
       const featureAttrs = this.findFeatureAttributesByIds(activeFeatureIds);
       const resolvedIds = new Set(selectedRequestedIds);
 
@@ -291,6 +299,7 @@ define(["backbone"], (Backbone) => {
 
       const registerTileWaiters = (layer) => {
         if (!this.isActiveSession(restoreSession)) return;
+        if (layer.get("status") === "error") return;
         if (typeof layer.waitForFeatureById !== "function") return;
 
         const selectedIds = (mapModel.getSelectedFeatures()?.models || []).map(
--- a/src/js/models/maps/LayerLoadingCoordinator.js
+++ b/src/js/models/maps/LayerLoadingCoordinator.js
@@ -24,8 +24,7 @@ define([], () => {
     if (layer.get("status") === "error") return false;
 
     return (
-      layer.get("status") === "loading" ||
-      layer.get("displayReady") === false
+      layer.get("status") === "loading" || layer.get("displayReady") === false
     );
   }
 
@@ -36,7 +35,9 @@ define([], () => {
    * @since 0.0.0
    */
   function getTrackedLoadingLayers(mapModel) {
-    return mapModel.getAllLayers().filter((layer) => isTrackedLayerLoading(layer));
+    return mapModel
+      .getAllLayers()
+      .filter((layer) => isTrackedLayerLoading(layer));
   }
 
   /**
@@ -67,7 +68,9 @@ define([], () => {
    */
   function getLoadingLayerLabel(layer) {
     const label = layer?.get("label");
-    return typeof label === "string" && label.trim().length ? label.trim() : null;
+    return typeof label === "string" && label.trim().length
+      ? label.trim()
+      : null;
   }
 
   /**
@@ -89,7 +92,10 @@ define([], () => {
    * @returns {string|null} The user-facing loading message.
    * @since 0.0.0
    */
-  function getLoadingLayersMessage(mapModel, loadingLayers = getTrackedLoadingLayers(mapModel)) {
+  function getLoadingLayersMessage(
+    mapModel,
+    loadingLayers = getTrackedLoadingLayers(mapModel),
+  ) {
     const labels = getLoadingLayerLabels(loadingLayers);
     if (!labels.length) {
       return mapModel.get("isLoadingLayers") ? "Loading layers" : null;
--- a/src/js/models/maps/LayerLoadingCoordinator.js
+++ b/src/js/models/maps/LayerLoadingCoordinator.js
@@ -1,0 +1,151 @@
+"use strict";
+
+define([], () => {
+  /**
+   * Determine whether a layer should participate in the aggregate loading state.
+   * Internal helper layers can opt out explicitly.
+   * @param {Backbone.Model|object} layer The layer model to check.
+   * @returns {boolean} Whether the layer should be tracked.
+   * @since 0.0.0
+   */
+  function shouldTrackLayerLoading(layer) {
+    return layer?.get("excludeFromLoadingState") !== true;
+  }
+
+  /**
+   * Return true when a tracked layer is enabled but not yet displayed.
+   * @param {Backbone.Model|object} layer The layer model to check.
+   * @returns {boolean} Whether the layer is still loading.
+   * @since 0.0.0
+   */
+  function isTrackedLayerLoading(layer) {
+    if (!shouldTrackLayerLoading(layer)) return false;
+    if (layer?.get("visible") !== true) return false;
+    if (layer.get("status") === "error") return false;
+
+    return (
+      layer.get("status") === "loading" ||
+      layer.get("displayReady") === false
+    );
+  }
+
+  /**
+   * Get visible tracked layers that are actively loading.
+   * @param {Backbone.Model} mapModel The parent map model.
+   * @returns {Array} The layers still loading into the map.
+   * @since 0.0.0
+   */
+  function getTrackedLoadingLayers(mapModel) {
+    return mapModel.getAllLayers().filter((layer) => isTrackedLayerLoading(layer));
+  }
+
+  /**
+   * Sync each layer's canonical loading flag to match current tracked loading state.
+   * @param {Backbone.Model} mapModel The parent map model.
+   * @param {Backbone.Model[]} [loadingLayers] Optional precomputed loading layers.
+   * @since 0.0.0
+   */
+  function syncTrackedLayerLoadingFlags(
+    mapModel,
+    loadingLayers = getTrackedLoadingLayers(mapModel),
+  ) {
+    const loadingLayerSet = new Set(loadingLayers);
+
+    mapModel.getAllLayers().forEach((layer) => {
+      if (!layer) return;
+      const isLoadingLayer = loadingLayerSet.has(layer);
+      if (layer.get("isLoadingLayer") === isLoadingLayer) return;
+      layer.set("isLoadingLayer", isLoadingLayer);
+    });
+  }
+
+  /**
+   * Get a user-facing label for a loading layer.
+   * @param {Backbone.Model|object} layer The layer model.
+   * @returns {string|null} The label to surface in the loading message.
+   * @since 0.0.0
+   */
+  function getLoadingLayerLabel(layer) {
+    const label = layer?.get("label");
+    return typeof label === "string" && label.trim().length ? label.trim() : null;
+  }
+
+  /**
+   * Get the distinct labels for layers contributing to the current loading state.
+   * @param {Backbone.Model[]} [loadingLayers] Optional precomputed loading layers.
+   * @returns {string[]} A deduplicated list of loading layer labels.
+   * @since 0.0.0
+   */
+  function getLoadingLayerLabels(loadingLayers) {
+    return loadingLayers
+      .map((layer) => getLoadingLayerLabel(layer))
+      .filter(Boolean);
+  }
+
+  /**
+   * Format the map-level loading message from the loading layer labels.
+   * @param {Backbone.Model} mapModel The parent map model.
+   * @param {Backbone.Model[]} [loadingLayers] Optional precomputed loading layers.
+   * @returns {string|null} The user-facing loading message.
+   * @since 0.0.0
+   */
+  function getLoadingLayersMessage(mapModel, loadingLayers = getTrackedLoadingLayers(mapModel)) {
+    const labels = getLoadingLayerLabels(loadingLayers);
+    if (!labels.length) {
+      return mapModel.get("isLoadingLayers") ? "Loading layers" : null;
+    }
+
+    if (labels.length === 1) {
+      return `Loading ${labels[0]}`;
+    }
+
+    if (labels.length === 2) {
+      return `Loading ${labels[0]} and ${labels[1]}`;
+    }
+
+    return `Loading ${labels[0]} and ${labels.length - 1} more layers`;
+  }
+
+  /**
+   * Sync the aggregate map loading indicator state.
+   * @param {Backbone.Model} mapModel The parent map model.
+   * @since 0.0.0
+   */
+  function updateLayerLoadingState(mapModel) {
+    const loadingLayers = getTrackedLoadingLayers(mapModel);
+    syncTrackedLayerLoadingFlags(mapModel, loadingLayers);
+
+    const hadLoadingState = mapModel.get("isLoadingLayers") === true;
+    const isLoadingLayers = loadingLayers.length > 0;
+    const loadingLayersMessage = isLoadingLayers
+      ? getLoadingLayersMessage(mapModel, loadingLayers) || "Loading layers"
+      : null;
+
+    if (
+      mapModel.get("isLoadingLayers") === isLoadingLayers &&
+      mapModel.get("loadingLayersMessage") === loadingLayersMessage
+    ) {
+      return;
+    }
+
+    mapModel.set({
+      isLoadingLayers,
+      loadingLayersMessage,
+    });
+
+    if (!hadLoadingState && isLoadingLayers && !loadingLayersMessage) {
+      mapModel.trigger("loading:started");
+    }
+  }
+
+  return {
+    shouldTrackLayerLoading,
+    isTrackedLayerLoading,
+    getTrackedLoadingLayers,
+    syncTrackedLayerLoadingFlags,
+    getLoadingLayerLabel,
+    getLoadingLayerLabels,
+    getLoadingLayersMessage,
+    updateLayerLoadingState,
+  };
+});
--- a/src/js/models/maps/Map.js
+++ b/src/js/models/maps/Map.js
@@ -442,6 +442,8 @@ define([
           activeVisualizationAction: null,
           activeVisualizationActionId: null,
           activeVisualizationUrl: null,
+          isLoadingLayers: false,
+          loadingLayersMessage: null,
         };
       },
 
@@ -542,6 +544,8 @@ define([
           mapModel: this,
         });
         this.featureRestoreSession = null;
+        this.loadingStateLayerGroups = [];
+        this.setUpLayerLoadingStateListeners();
         this.setUpUrlStateListeners();
         this.applyRestoreState();
       },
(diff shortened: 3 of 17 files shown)


src/js/models/maps/Map.js
```diff
@@ -601,8 +617,78 @@ define([
...
+              this.handleLayerVisibilityChange,
+            );
+          });
+        }
+
+        this.loadingStateLayers = this.getAllLayers();
+        this.listenTo(
+          this,
+          "change:layers change:layerCategories",
+          this.setUpLayerLoadingStateListeners,
```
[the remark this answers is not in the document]
DarianGill: Addressed in e9ef61aad1e2983710c3b8bb196aa9a500de6715

src/js/models/maps/Map.js
```diff
@@ -601,8 +617,78 @@ define([
...
+       * Hidden layers should not keep map loading state active.
+       * @since 0.0.0
+       */
+      handleLayerVisibilityChange() {
+        const activeFeatureIds = this.get("restoreState")?.activeFeatureIds;
+        if (
+          this.shouldSyncUrlState() &&
+          isNonEmptyArray(activeFeatureIds) &&
+          this.featureRestoreSession
+        ) {
```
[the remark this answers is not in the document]
DarianGill: Addressed in 6d0ac9df8e2df86194fefc1cfdc0d571ac6d4b4e patching this and ensuring toggling off a layer with a selected feature consistently clears that feature from selection even if the feature is mid-restore.

src/js/models/maps/Map.js
```diff
@@ -601,8 +617,78 @@ define([
...
+          this.setUpLayerLoadingStateListeners,
+        );
+
+        if (this.loadingStateLayers?.length) {
+          this.loadingStateLayers.forEach((layer) => {
+            this.stopListening(
+              layer,
+              "change:status change:displayReady change:label",
+              () => LayerLoadingCoordinator.updateLayerLoadingState(this),
+            );
```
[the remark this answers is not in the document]
DarianGill: Addressed in https://github.com/NCEAS/metacatui/commit/e9ef61aad1e2983710c3b8bb196aa9a500de6715

src/js/models/maps/ActiveFeatureRestoreController.js
```diff
@@ -0,0 +1,390 @@
...
+      });
+
+      const unresolvedIds = activeFeatureIds.filter(
+        (id) => !resolvedIds.has(id),
+      );
+      const restoreKey = activeFeatureIds.join(",");
+      const searchableLayerKey = allSearchableLayers
+        .map((layer) => layer.get("layerId") || layer.cid || "")
+        .join(",");
+      const restoreScopeKey = `${restoreKey}|${searchableLayerKey}`;
```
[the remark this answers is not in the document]
DarianGill: Addressed in f7d0cf3ed9bdb7d2ce5c76b67618e76d7a226161

src/js/views/maps/CesiumWidgetView.js
```diff
@@ -1753,6 +1779,16 @@ define([
       addImagery(cesiumModel) {
         this.scene.imageryLayers.add(cesiumModel);
         this.sortImagery();
+        const mapAsset = cesiumModel?.mapAssetModel;
+        if (mapAsset) {
+          const removeListener = this.scene.postRender.addEventListener(() => {
+            removeListener();
+            if (mapAsset.get("displayReady") !== true) {
+              mapAsset.set("displayReady", true);
+            }
```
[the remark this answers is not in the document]
DarianGill: Addressed in 1c403d6b30776fd7f01df71add0474ef41b3dc7c where I added a shared display-ready lifecycle API to the base asset model in MapAsset.js so that each layer type reliably has fetched/decoded at least one drawable tile before makring displayReady, instead of just using postRender.

src/js/views/maps/MapView.js
```diff
@@ -96,6 +102,63 @@ define([
...
+          const loadingIndicator = indicator;
+          this.loadingIndicatorTimer = setTimeout(() => {
+            this.loadingIndicatorTimer = null;
+            if (this.model.get("isLoadingLayers") !== true) {
+              return;
+            }
+
+            messageTextEl.textContent =
+              this.model.get("loadingLayersMessage") || "Loading layers";
+            loadingIndicator.hidden = false;
```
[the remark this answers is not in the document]
DarianGill: Addressed in 105f70ecf4c902611778e0ff8a54ae98b2ab7c95

src/js/models/maps/Map.js
```diff
@@ -613,6 +647,115 @@ define([
...
+       */
+      handleLayerVisibilityChange(layer, visible) {
+        const activeFeatureIds = this.get("restoreState")?.activeFeatureIds;
+
+        if (
+          visible === false &&
+          this.shouldSyncUrlState() &&
+          isNonEmptyArray(activeFeatureIds)
+        ) {
+          this.clearFeatureRestoreSession();
```
robyngit, line 743:
```suggestion
```

Clearing the whole restore session here drops IDs that the app has not found yet, e.g. when the feature is still loading in another visible layer.

For example, if A has loaded, B is still pending, and unrelated layer C is hidden, B is removed from the URL and restore state. The app also stops listening for the feature to finish loading.

robyngit (commented):
I'm happy to see a much needed loading progress bar! Makes what's happening so much clearer. I like the addition of showing the loading spinner in the layer list and naming the exact layers that are still loading in the progress bar. Smart!

My initial feeling was that the progress bar took up too much space, or maybe just felt too prominent. Did you experiment with any other placements or layouts? This is just my personal opinion, I don't think it's a blocker.

The larger concern is that `displayReady` means "this layer has produced one visible tile," rather than "this layer has no pending work."

This can leave the loading indicator stuck when a visible layer has no tiles in the current view. For example, I found the "Retrogressive Thaw Slumps (RTS) Outlines by DARTS" layer showed the loading indicator endlessly when no tiles were visible in the view, but none were loading in the network tab.

The opposite happens after the first tile appears. The tracking stops, but tiles could still be loading. Also later tile requests while panning are not shown as loading.

We need a fix that handles both imagery and 3D tiles correctly. Some sort of signal for pending requests or processing for each layer. If this PR is only meant to cover initial layer setup, maybe a smaller fix would be to stop at provider or tileset readiness and add a fallback when no tile appears in the view. That would prevent the permanent loading indicator, but it wouldn't cover later loading while panning.

Do you think this is feasible? Maybe the complete loading lifecycle needs a separate issue and some investigation?

src/js/models/maps/Map.js
DarianGill: Yeah I couldn't find a quick way to deal with reconciling the feature/layer restore with mid-restore user interaction that was potentially interruptive, but I agree it's messy. It seemed less clunky than just locking the UI but there has to be a better option. Maybe now that the feature id is encoded with a layer id I can gate this better and only cancel the restoration if the layer containing the feature is toggled.

DarianGill:
Thanks for the feedback @robyngit. I get where you're coming from with prominence, but I couldn't find a better placement without sacrificing specificity or overlapping existing ui widgets/information. It took some finagling to ensure this indicator wouldn't overlap the scale bar in the upper right for example, and when I didn't add text to the widget I was worried the spinners in the layer menu may not be noticeable. Do you have a mockup or example you could point me to for a more minimal solution? The [one you mentioned in the issue](https://nationalmap.gov.au/) unfortunately seems to have been decommissioned. 

I'm glad you picked up on the displayReady limitations, as I meant to include a paragraph about that in the PR description but now see I neglected to do so. I believe you're right that with the current implementation for a 3D tiles layer, you may have to move the map (or otherwise get any tile from that layer into view) for the loading widget to dismiss. I wanted to talk this through with you, because I thought you'd be more familiar with the lifecycle and was hoping that there was an existing event I could listen for instead. It felt a bit beyond the scope of the PR to dive into changing the layer lifecycle and was quickly growing complicated trying to get the feature and layer restoration to work well with this without blocking user interaction with the map until it's fully loaded. If theres a better event I can listen for that'd be music to my ears.

DarianGill:
@robyngit Since I couldn't find a better flag than displayReady, and you've been busy with other big things, I made tilesLoading in [6b905e3](https://github.com/NCEAS/metacatui/pull/2888/commits/6b905e30561b7e231e47634f7f9387ec9ef13142). It keys into that pending work you mentioned for both 3d tilesets and imagery:
- Cesium3DTileset: driven by the tileset's own `loadProgress` event
  (numberOfPendingRequests / numberOfTilesProcessing).
- CesiumImagery: driven by wrapping the imagery provider's
  requestImage so outstanding tile requests are counted directly
  from the promises Cesium creates.

I also renamed MapAsset's startDisplayReadyTracking/stopDisplayReadyTracking
to startLoadingStateTracking/stopLoadingStateTracking, and keep their
listeners alive for the layer's lifetime in the scene instead of
tearing down after the first tile.

LayerLoadingCoordinator.isTrackedLayerLoading() now prefers
`tilesLoading` when available too, falling back to the previous
`displayReady === false` check for layer types that don't track it
(e.g. vector data). And the Map's loading-state listeners now also watch
change:tilesLoading.

This seems to account well for all the zooming and panning and out of view potentialities, but does make the loading indicator flash quickly in and out, highlighting the prominence challenges you mentioned earlier. That can kinda be managed by adjusting LOADING_INDICATOR_DELAY_MS, but only so much. I could see making the indicator just a muted loading bar at the top of the map like the [Copernicus browser](https://browser.dataspace.copernicus.eu/?zoom=6&lat=47.7984&lng=-117.13623&themeId=DEFAULT-THEME&visualizationUrl=U2FsdGVkX18s2O%2Bkx6k%2BtnA%2BnG1Si5Xb6QkbLEyfs2O8eh1%2FKNDtlyZ4hfCQyYXcXIRjekm5X3Be1WVApFTTKHbGPv5DpbkMsdn7HN9K30%2FRsJVTSO6v1oYDhE3UjML5&datasetId=S2_L2A_CDAS&fromTime=2023-02-07T00%3A00%3A00.000Z&toTime=2023-02-07T23%3A59%3A59.999Z&layerId=3-WILDFIRES&demSource3D=%22MAPZEN%22&cloudCoverage=10&dateMode=SINGLE) for landsat imagery has, but I'm concerned folks may not know what they're waiting for if we take away the text in the widget. There would still be the spinners in the layer menu, but those are collapsible and could also be hidden by the viewfinder. 

Please take a look and lemme know what you think, I could definitely be convinced.

DarianGill:
@robyngit Oh also, working on this has highlighted to me how many delegators and and utility/unrelated functions have found their way into map.js. I was beginning a refactor that offered better separation of concerns, but stopped because I think that's best saved for a pr dedicated to issue https://github.com/NCEAS/metacatui/issues/2870. I intend to start that work next week as soon as we can get these two PRs into develop, as I don't want the scope to drift anymore than it already has between them and I'd like to avoid further merge conflicts if possible. That work will probably tidy up map.js to just listen for and handle events, putting the actual logic into layerLoadingCoordinator, activeFeatureRestoreController, and any other related modules. I don't think we should make a release with these until I can tidy up a bit, if it's ok with you.

src/js/views/maps/MapView.js
```diff
@@ -318,6 +319,7 @@ define([
        * @since 2.27.0
        */
       onClose() {
+        this.clearLayerLoadingIndicatorTimer();
```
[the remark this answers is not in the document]
DarianGill: Fixed in c8aedfe9985c2c3f67dbae0e411dfd42a592e55e

src/js/models/maps/LayerLoadingCoordinator.js
```diff
@@ -0,0 +1,211 @@
...
+
+    if (labels.length === 1) {
+      return `Loading ${labels[0]}`;
+    }
+
+    // As soon as more than one layer is loading, switch to a compact "+n other
+    // layers" form so the message doesn't keep growing with every extra label.
+    const otherCount = labels.length - 1;
+    const otherWord = otherCount === 1 ? "layer" : "layers";
+    return `Loading ${labels[0]} and ${otherCount} other ${otherWord}`;
```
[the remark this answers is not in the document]
DarianGill: Resolved in be5a705c6637773b64496c503decb9d7846e8cd3

src/js/models/maps/assets/Cesium3DTileset.js
```diff
@@ -446,6 +447,98 @@ define([
...
+      startLoadingStateTracking: function () {
+        const model = this;
+        const cesiumModel = model.get("cesiumModel");
+        this.stopLoadingStateTracking();
+
+        if (!cesiumModel) {
+          return;
+        }
+
+        model.set("tilesLoading", true);
```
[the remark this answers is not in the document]
DarianGill: Done in 9317504d24b710a33200b630649e0fd7bf30d8b0

src/js/models/maps/assets/CesiumImagery.js
```diff
@@ -317,6 +319,192 @@ define([
...
+        const scene = context?.scene;
+        const model = this;
+        this.stopLoadingStateTracking();
+
+        const imageryLayer = this.get("cesiumModel");
+        if (!scene || !imageryLayer) {
+          return;
+        }
+
+        model.set("tilesLoading", true);
```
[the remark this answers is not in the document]
DarianGill: Done in https://github.com/NCEAS/metacatui/commit/9317504d24b710a33200b630649e0fd7bf30d8b0

src/js/views/maps/MapWidgetContainerView.js
```diff
@@ -63,46 +63,47 @@ define([
...
           );
           return;
         }
-        const scaleBar = new ScaleBarView({
-          // el: this.el,
+        const statusBar = new MapStatusBarView({
+          model: this.model,
           scaleModel: interactions.get("scale"),
           pointModel: interactions.get("mousePosition"),
         });
```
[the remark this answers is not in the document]
DarianGill: Fixed in 39fd897b1f2987dc5740c44e92b17833c87c8073

src/js/models/maps/Map.js
```diff
@@ -663,6 +697,122 @@ define([
...
+              layer,
+            );
+          if (removedRestoreFeatures) {
+            const removedSelectedFeatures =
+              this.clearSelectedFeaturesForLayer(layer);
+            if (!removedSelectedFeatures) {
+              this.syncSelectedFeaturesToUrl();
+            }
+            this.handleLayerLoadingStateChange();
+            return;
```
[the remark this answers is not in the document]
DarianGill: Done in b8b2c7c7c96398f853d773849a1dffc399dfaddd

src/js/models/maps/assets/MapAsset.js
```diff
@@ -751,11 +759,33 @@ define([
        * @since 2.21.0
        */
       resetStatus() {
-        const defaults = this.defaults();
-        this.set("status", defaults.status);
-        this.set("statusDetails", defaults.statusDetails);
+        this.stopLoadingStateTracking();
+        this.set("status", "loading");
+        this.set("statusDetails", null);
+        this.set("displayReady", false);
```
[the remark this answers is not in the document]
DarianGill: Done in 0d5788a6b1b5570762ecff3881c790d076e6d293

src/js/models/maps/Map.js
DarianGill: Think I managed to do that.

src/js/models/maps/ActiveFeatureRestoreController.js
```diff
@@ -273,26 +304,91 @@ define(["backbone"], (Backbone) => {
...
         requestedFeatures: normalizedFeatures.slice(),
       });
+
+      // Ensure restores have a bounded lifetime so unresolved waits do not
+      // persist forever when no matching feature ever arrives.
+      const timeoutMs = this.mapModel.get("featureRestoreTimeoutMs") || 15000;
+      const timeoutId = setTimeout(() => {
+        if (!this.isActiveSession(session)) return;
+        this.clearSession();
+        LayerLoadingCoordinator.updateLayerLoadingState(this.mapModel);
```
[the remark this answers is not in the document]
DarianGill: Done in e545bd8f41e57cec8da72f3d84595af9bb7a725d

DarianGill:
Ok, this became quite the behemoth, but I think it's in a robust and functional spot ready for your review now @robyngit  :) Don't hesitate to lemme know if that's not the case, or if there are other obvious improvements I should make to make our lives easier in future.

src/js/models/maps/LayerLoadingCoordinator.js
```diff
@@ -0,0 +1,211 @@
...
+   * already renders as HTML, see `LayerItemView`/`layer-item.html`) so the
+   * loading message reads the same as the layer menu.
+   * @param {Backbone.Model|object} layer The layer model.
+   * @returns {string|null} The label to surface in the loading message.
+   * @since 0.0.0
+   */
+  function getLoadingLayerLabel(layer) {
+    const label = layer?.get("label");
+    if (typeof label !== "string" || !label.trim().length) return null;
+    return truncateLabel(label.trim());
```
robyngit: nitpicky but anything with HTML DOM manipulation belongs in the view rather than the model. Same with truncateLabel, above.

src/js/models/maps/Map.js
```diff
@@ -421,6 +425,7 @@ define([
             {
               type: "OpenStreetMapImageryProvider",
               label: "Base layer",
+              excludeFromLoadingState: true,
```
robyngit: Why is this one excluded?

src/js/views/maps/LayerLoadingIndicatorView.js
```diff
@@ -0,0 +1,63 @@
...
+       * The primary HTML template for this view
+       * @type {Underscore.template}
+       */
+      template: _.template(template),
+
+      /** @inheritdoc */
+      render() {
+        this.$el.html(this.template());
+        this.el.setAttribute("aria-live", "polite");
+        this.el.setAttribute("aria-atomic", "true");
```
robyngit: Great!

src/js/views/maps/LayerLoadingIndicatorView.js
```diff
@@ -0,0 +1,63 @@
...
+        this.messageEl = this.el.querySelector(".map-status-bar__loading-text");
```
robyngit: Need to add this to CONTRIBUTING.md, but during the Google fellows days, we decided on a convention where class names (and other strings, like user-facing messages) should generally get defined as a const at the top of the file (search for `const CLASS_NAMES =` for examples). Class name definition is wildly inconsistent across the code base, but we should generally move toward some sort of convention with how we do this.

Actually, I see you did this for other view files, so probably just an oversight here!

src/js/views/maps/LayerLoadingIndicatorView.js
```diff
@@ -0,0 +1,63 @@
...
+  const LayerLoadingIndicatorView = Backbone.View.extend(
+    /** @lends LayerLoadingIndicatorView.prototype */ {
+      /** @inheritdoc */
+      className: "map-status-bar__loading-row",
+
+      /**
+       * The primary HTML template for this view
+       * @type {Underscore.template}
+       */
+      template: _.template(template),
```
robyngit: Trying to migrate away from underscore, and the template you define doesn't have any templating needed, so you might as well just make template a static string here. (or a fn that takes the string and returns the DOM)

src/js/views/maps/MapStatusBarView.js
```diff
@@ -0,0 +1,282 @@
...
+   * (e.g. several layers finishing/starting within milliseconds of each other),
+   * so updates are throttled to this rate rather than applied as soon as they
+   * happen, which would make the message flicker between different layer
+   * names/counts.
+   * @type {number}
+   */
+  const MESSAGE_UPDATE_INTERVAL_MS = 400;
+
+  const CLASS_NAMES = {
+    expanded: "map-status-bar--loading-expanded",
```
robyngit: Yes, like this

src/js/views/maps/MapStatusBarView.js
```diff
@@ -0,0 +1,282 @@
...
+   * @classcategory Views/Maps
+   * @name MapStatusBarView
+   * @augments Backbone.View
+   * @since 0.0.0
+   * @constructs
+   */
+  const MapStatusBarView = Backbone.View.extend(
+    /** @lends MapStatusBarView.prototype */ {
+      /** @inheritdoc */
+      className: "map-status-bar",
```
robyngit: nitpick: Might as well add this to CLASS_NAMES. Even better, define the BASE_CLASS and use it in CLASS_NAMES like in search input view, for example:

```js
  const BASE_CLASS = "search-input";
  const CLASS_NAMES = {
    searchButton: `${BASE_CLASS}__search-button`,
    searchButtonActive: `${BASE_CLASS}__search-button--active`,
    cancelButton: `${BASE_CLASS}__cancel-button`,
    cancelButtonContainer: `${BASE_CLASS}__cancel-button-container`,
    input: `${BASE_CLASS}__input`,
    inputField: `${BASE_CLASS}__field`,
    errorInput: `${BASE_CLASS}__error-input`,
    errorText: `${BASE_CLASS}__error-text`,
  };
```

src/js/views/maps/MapStatusBarView.js
```diff
@@ -0,0 +1,282 @@
...
+        this.model = options.model;
+        this.scaleModel = options.scaleModel;
+        this.pointModel = options.pointModel;
+        this.revealTimer = null;
+        this.collapseTimer = null;
+        this.messageUpdateTimer = null;
+        this.lastMessageUpdateAt = 0;
+        this.pendingMessage = null;
+        this.expandedAt = null;
+        this.isCurrentlyLoading = false;
```
robyngit: Convention is to add these as properties on the view directly (outside of initialize) and give each a little JS docs def. Just another nitpick :)

src/js/views/maps/MapStatusBarView.js
```diff
@@ -0,0 +1,282 @@
...
+       * @param {object} options The options for this view.
+       * @param {Map} options.model The map model that provides the aggregate
+       * `isLoadingLayers`/`loadingLayersMessage` loading state.
+       * @param {GeoScale} options.scaleModel Passed through to the ScaleBarView.
+       * @param {GeoPoint} options.pointModel Passed through to the ScaleBarView.
+       */
+      initialize(options = {}) {
+        this.model = options.model;
+        this.scaleModel = options.scaleModel;
+        this.pointModel = options.pointModel;
```
robyngit: Do these need defaults or should the view just throw if not provided?

src/js/models/maps/assets/CesiumImagery.js
```diff
@@ -317,6 +319,192 @@ define([
...
+      isDisplayReadyInScene: function (scene) {
+        const imageryLayer = this.get("cesiumModel");
+        if (!imageryLayer || !scene || !scene.imageryLayers) {
+          return false;
+        }
+        if (!imageryLayer.show || !scene.imageryLayers.contains(imageryLayer)) {
+          return false;
+        }
+
+        const imageryCache = imageryLayer["_imageryCache"];
```
robyngit: If this is to avoid the "no-underscore-dangle" lint rule, don't worry, I removed that rule in the develop branch.

I know this is the only way to do it, but if/when we update Cesium, we'll have to carefully check this still works since it's a private method and no guarantees it won't change.

src/js/models/maps/assets/CesiumImagery.js
```diff
@@ -317,6 +319,192 @@ define([
...
+      ) {
+        if (!provider || typeof provider.requestImage !== "function") {
+          return function () {};
+        }
+
+        const originalRequestImage = provider.requestImage.bind(provider);
+        let pendingCount = 0;
+        let isUnwrapped = false;
+
+        provider.requestImage = function (x, y, level, request) {
```
robyngit: Does this mean we can spy on when Cesium is loading different (z) levels of imagery?? This could be very useful for other features!

src/js/views/maps/MapWidgetContainerView.js
```diff
@@ -38,7 +38,7 @@ define([
         this.renderLegendContainer();
 
         if (this.model.get("showScaleBar")) {
-          this.renderScaleBar();
+          this.renderStatusBar();
```
robyngit: showScaleBar means show/hide the scale bar, but this now also blocks rendering the loading indicator.

Could we render the loading row independently of the scale bar? (and if needed in the future, add a mapConfig option for showLoadingIndicator)?

To do that, we'd need to remove the condition here and move it to MapStatusBarView, and make it only block the rendering of the scaleBarView.

If it's too much surgery for this PR, we can always address this later. Not a merge blocker.

src/js/models/maps/assets/Cesium3DTileset.js
```diff
@@ -446,6 +448,98 @@ define([
...
+          if (isCanceled) {
+            return;
+          }
+          if (model.get("cesiumModel") !== cesiumModel) {
+            cleanupListeners();
+            return;
+          }
+          model.set(
+            "tilesLoading",
+            numberOfPendingRequests > 0 || numberOfTilesProcessing > 0,
```
robyngit: Why use `numberOfPendingRequests > 0 || numberOfTilesProcessing > 0` and not `!cesiumModel.tilesLoaded`? What's the difference?

src/js/models/maps/LayerLoadingCoordinator.js
```diff
@@ -0,0 +1,211 @@
...
+
+    mapModel.set({
+      isLoadingLayers,
+      loadingLayersMessage,
+    });
+
+    if (!hadLoadingState && isLoadingLayers && !loadingLayersMessage) {
+      mapModel.trigger("loading:started");
+    }
+  }
```
robyngit:
```suggestion
```

No listener for this event

src/js/models/maps/ActiveFeatureRestoreController.js
```diff
@@ -183,6 +187,21 @@ define(["backbone"], (Backbone) => {
...
+   * @param {string[]} layerIds Searchable visible layer ids.
+   * @returns {string} Stable scope key.
+   * @since 0.0.0
+   */
+  function serializeRestoreScopeKey(featureKeys, layerIds) {
+    return JSON.stringify({
+      featureKeys: [...new Set(featureKeys)].sort(),
+      layerIds: [...new Set(layerIds)].sort(),
+    });
+  }
```
robyngit: Nice

robyngit (approved):
This is so well done, @DarianGill. You call this PR a "behemoth" but actually, it was a pleasure to review. Very readable and structured approach, and you thought of all the details. More importantly, the indicator works so well! I already hate using the prod PDG portal that doesn't have it yet. The expansion/collapse behaviour is smooth, and it now draws the appropriate amount of attention (not too flashy or in the way). It's small in terms of screen real estate, but huge in the impact it will have on making the map just so much more usable. Thanks!!

Added some comments, questions, suggestions, but none are merge blocking.

src/js/views/maps/LayerLoadingIndicatorView.js
DarianGill: Oh that's right! I forgot about that and noticed underscore in another view so replicated here. I'll put in the static string approach like I did in previous PRs.

src/js/views/maps/LayerLoadingIndicatorView.js
DarianGill: Oop yes! I meant to 🤦 Thanks for the catch!

src/js/models/maps/Map.js
DarianGill: Ah yes I think this was just a holdover that snuck through from some dev testing. The layers are meant to be piecemeal excludable, but I think the only one we want to exclude is the draw your polygon one.

src/js/models/maps/LayerLoadingCoordinator.js
DarianGill: Ah, fair. This function used to be just string manipulation, but I opted for html handling because special characters like the subscript in CO₂ would render as raw literal html tags <sub>2</sub> instead of characters. It does feel more like part of the "business logic" than rendering to me especially because it's partially pulling the values from the model, but truncateLabel certainly manipulates the dom. @robyngit You'd want all of these lines moved to LayerLoadingIndicatorView?

```
  /**
   * Labels longer than this are truncated with an ellipsis so a single very long
   * layer name can't dominate the loading message width.
   * @type {number}
   */
  const MAX_LABEL_LENGTH = 50;

  /**
   * Truncate a layer label's visible text to {@link MAX_LABEL_LENGTH} characters,
   * appending an ellipsis when truncated. Labels can contain HTML markup (e.g.
   * `<sub>`/`<sup>` tags for chemical/mathematical notation, rendered as real
   * HTML by the layer menu, see `layer-item.html`); truncating is done by
   * parsing the label with the browser's own HTML parser and trimming text
   * nodes, so any markup that survives stays well-formed instead of being cut
   * off mid-tag.
   * @param {string} label The label to truncate, which may contain HTML markup.
   * @returns {string} The truncated label, as an HTML string.
   * @since 0.0.0
   */
  function truncateLabel(label) {
    const container = document.createElement("div");
    container.innerHTML = label;
    if (container.textContent.length <= MAX_LABEL_LENGTH) return label;

    let remaining = MAX_LABEL_LENGTH;
    const trimNode = (node) => {
      if (remaining <= 0) {
        node.remove();
        return;
      }
      if (node.nodeType === Node.TEXT_NODE) {
        const truncated = node.textContent.slice(0, remaining);
        // eslint-disable-next-line no-param-reassign
        node.textContent = truncated;
        remaining -= truncated.length;
        return;
      }
      Array.from(node.childNodes).forEach(trimNode);
    };
    Array.from(container.childNodes).forEach(trimNode);

    return `${container.innerHTML}…`;
  }

  /**
   * Get a user-facing label for a loading layer. The label is returned as-is
   * (including any HTML markup, like the `<sub>`/`<sup>` tags the layer menu
   * already renders as HTML, see `LayerItemView`/`layer-item.html`) so the
   * loading message reads the same as the layer menu.
   * @param {Backbone.Model|object} layer The layer model.
   * @returns {string|null} The label to surface in the loading message.
   * @since 0.0.0
   */
  function getLoadingLayerLabel(layer) {
    const label = layer?.get("label");
    if (typeof label !== "string" || !label.trim().length) return null;
    return truncateLabel(label.trim());
  }

  /**
   * Get the distinct labels for layers contributing to the current loading state.
   * @param {Backbone.Model[]} [loadingLayers] Optional precomputed loading layers.
   * @returns {string[]} A deduplicated list of loading layer labels.
   * @since 0.0.0
   */
  function getLoadingLayerLabels(loadingLayers) {
    return loadingLayers
      .map((layer) => getLoadingLayerLabel(layer))
      .filter(Boolean);
  }
```

src/js/models/maps/Map.js
DarianGill: Removed in 59474aa04a39354b4a9f00cd0119da21da1e4ad8

src/js/views/maps/LayerLoadingIndicatorView.js
DarianGill: Done in 0c181659788a23a5edaad737e63bffdc16443bb8

src/js/views/maps/LayerLoadingIndicatorView.js
DarianGill: Done in 0c181659788a23a5edaad737e63bffdc16443bb8

src/js/views/maps/MapStatusBarView.js
DarianGill: I just pulled the defaults from ScaleBarView which shared the dependencies. Are there other places that throw in the maps folder if they aren't provided? I think the this.ListenTo(this.model...) call would throw preventing rendering from completing so it wouldn't be a silent failure anyway.

src/js/views/maps/MapStatusBarView.js
DarianGill: Done in 4905a5e30af97c126379122d99caa6241853ea90

src/js/views/maps/MapStatusBarView.js
DarianGill: Done in 4905a5e30af97c126379122d99caa6241853ea90

src/js/models/maps/LayerLoadingCoordinator.js
DarianGill: Removed this and some unused CSS classes in f3f56220ead2905b0514057c45670caf8852025f

src/js/models/maps/assets/Cesium3DTileset.js
DarianGill: This was part of the move to an event-driven "still working" signal instead of a polled "done for this view" flag. I think tilesLoaded only updates when Cesium actually renders a frame for that tileset, so it required polling every frame. Also, I think it's scoped to the tiles selected for the current camera view, so during camera movement it can briefly flip false/true/false as the traversal re-evaluates each frame. The pending request tracking in contrast was more stable in my experience and better accommodates background loading (e.g., sibling/foveated preloading). I could be missing something though! Am I mistaken?

src/js/models/maps/assets/CesiumImagery.js
DarianGill: Sweet! Back to the dot syntax as of c0ad2e61218f78d9bcac2ca9d888fca958f90eaf

Issue #1995, opened 2022-03-23
Cesium page creator in portal builder


laurenwalker:
The ADC data team might start adding Cesium to portals that have geojson layers, so we should add a simple interface for adding Cesium pages to portals.

PR #2893, opened 2026-09-03
Map Editor in Portals - Part 1 - Model/View cleanup

robyngit:
This is part 1 of feature #1995, adding a map editor in the portal editor.

Part 1 cleans up the five of the models and views that will be modified while working on this feature:

`PortalModel.js`
`PortalVizSectionModel.js`
`PortEditorMdSectionView.js`
`PortEditorSectionView.js`
`PortEditorSectionsView.js`

Clean up includes fixing all eslint errors, adding some basic tests, adding and fixing jsdocs, replacing underscore methods with native JS methods, removing unused dependencies, and ensuring the views return the view itself from the render methods to be consistent with the rest of MetacatUI.


src/js/models/portals/PortalModel.js
```diff
@@ -262,11 +265,8 @@ define([
         }
 
         // Can't get a random image if none are configured
-        if (!defaultImageIds) {
-          console.log(
-            "Can't set a default image on new markdown sections because there are no default image IDs set. Check portalDefaults.sectionImageIdentifiers in the config file.",
-          );
-          return;
+        if (!defaultImageIds || !defaultImageIds.length) {
+          return undefined;
```
DarianGill: Might it still be beneficial to log something if this returns undefined or an empty string?

src/js/common/ValueUtilities.js
```diff
@@ -283,7 +274,7 @@ define(["md5", "uuid", "common/ErrorUtilities"], (
      */
     listify(value) {
       if (value === undefined || value === null) return [];
-      if (Array.isArray(value)) return value;
+      if (Array.isArray(value)) return [...value];
```
DarianGill: Just making sure, nothing relies on this returning the same array instance, right? Looks like this change would always return a copy if value is already an array.

src/js/models/portals/PortalModel.js
```diff
@@ -218,28 +215,28 @@ define([
...
+       * app config, select an image to use for a portal section. The function
+       * will not return the same image until all the images have been returned
+       * at least once. If an image would return a 404 error, it is skipped. If
+       * all images give 404s, an empty string is returned.
+       * @returns {(PortalImage|string|undefined)} A portal image model, an
+       * empty string when configured images are unavailable, or undefined when
+       * no images are configured.
        */
-      getRandomSectionImage: function () {
+      getRandomSectionImage() {
```
DarianGill: Might this deserve its own unit test? Seems a little complicated and like we'd want to enforce the partial or complete 404 behaviors.

src/js/models/CollectionModel.js
```diff
@@ -246,7 +246,7 @@ define([
        * @param {XMLDocument} response - The XMLDocument returned from the fetch() AJAX call
        * @return {JSON} The result of the parsed XML, in JSON. To be set directly on the model.
        */
-      parse: function (json) {
+      parse: function (response) {
         //Start the empty JSON object
         var modelJSON = {},
```
DarianGill: I take it you're not cleaning up the linting in this one?

DarianGill (approved):
So happy to not be seeing red when I open these files :D 

Can confirm: all 3003 tests pass for me, the linting for the 5 js files (not Collection Model.js) now passes, and I can noodle around on your map editor test portal without issues. The diffs were pretty large for me to reliably absorb what's happening, but the changes you described in the PR and implemented make sense to me. 

One thing I noticed is that  `npm run format` makes some additional changes to CITATION, codemeta.json, CONTRIBUTING.md, PortalListView.js, and AccessRule.spec.js that aren't in this branch. I also left some questions for potential additional improvements, but you needn't necessarily do em. Seems you're already on the path of implementation with part2.

src/js/models/CollectionModel.js
robyngit: Wasn't going to since it's out of scope, but so are these three bugs I decided to fix in this branch! So yes, might as well clean collection model up too. Good point!

src/js/common/ValueUtilities.js
robyngit: No, nothing relies on the same instance getting returned, made sure to check!

src/js/models/portals/PortalModel.js
robyngit: It was never that useful of a console log since some repositories are configured to start with default images on markdown pages and others aren't, it's not really an error.

Example, dataone:

Image

vs. ADC:

Image

src/js/models/CollectionModel.js
robyngit: Done! ✅

src/js/models/portals/PortalModel.js
robyngit: Good point. Done!

robyngit:
@DarianGill Good call on the formatting, I'll run `npm run format` in develop and make sure to format my PRs going forward! I think I've address the remainder of your comments

